PR #427, opened 2026-04-15
Security: Path traversal allows arbitrary file download outside temporary extraction directory

tomaioo:
## Summary

Security: Path traversal allows arbitrary file download outside temporary extraction directory

## Problem

**Severity**: `High` | **File**: `app/api/archive_download.py:L33`

The download helper constructs `extracted_file_path` using `os.path.join(temp_dir, file_path.lstrip('/'))` and then serves it if it exists. This does not prevent `..` path segments, so a crafted `file_path` like `../../etc/passwd` can escape `temp_dir` and potentially expose arbitrary server files via `FileResponse`.

## Solution

Normalize and validate the resolved path before serving: use `resolved = os.path.realpath(os.path.join(temp_dir, file_path))` and ensure `resolved.startswith(os.path.realpath(temp_dir) + os.sep)`. Reject any path that escapes the extraction directory. Consider using `pathlib.Path.resolve()` with strict parent checks.

## Changes

- `app/api/archive_download.py` (modified)


karanhudia:
@tomaioo Ruff format check seems to be failing.

PR #708, opened 2026-06-27
feat(repositories): allow ssh:// repositories for managed agents

ioanalytica:
## Description

Agent-owned repositories are rejected for any `ssh://` path, a
`connection_id`, or `execution_target=ssh`
(`_reject_agent_repository_ssh_target`). However, the **execution path
already fully supports ssh:// for agents**:

- `agent/borg_ui_agent/backup.py` (`BackupCreatePayload.build_command`)
  uses the `borg` (v1) binary for `borg_version == 1`, passes the
  repository path verbatim, and appends `--remote-path` when set.
- `app/services/repository_executor.py` puts `path`, `borg_version` and
  `remote_path` into both the backup and repository-operation payloads.
- The runtime gate `validate_agent_backup_repository` does not inspect the
  path scheme at all.

So an agent can already run `borg -r ssh://… --remote-path=borg-1.4
create/check` over its own mounted SSH key — only the create/import
**validation guard** prevented configuring it. In practice this blocks the
common, fast setup of backing up node-local data to a **Hetzner Storage
Box via server-side borg-1.4** (ssh), forcing borg2+sftp (dumb storage,
much slower) instead.

### Change
Narrow the guard to reject only a **server-managed `connection_id`** (an
`ssh_connections` row that only the server can use). `ssh://` paths and
`execution_target=ssh` are now accepted for agent repositories; the agent
supplies its own SSH credentials.

### Notes
- The two existing tests assert `connection_id` rejection and remain green.
- Added `test_create_agent_repository_accepts_ssh_path`.
- `_create_agent_repository_record` still records `repository_type="local"`
  for agent repos; that's metadata (the agent backs up via `path`), so it's
  left unchanged — happy to derive it from the scheme if you'd prefer.

## Type of Change
- [x] New feature (non-breaking)

## Testing
- `tests/unit/test_api_repositories.py`

## Summary by CodeRabbit

* **Bug Fixes**
  * Agent-managed repository creation now supports `ssh://...` paths with `execution_target="ssh"` when no server-side connection is specified.
  * Repository details are now persisted correctly for agent-based SSH setups, including the SSH URL and related remote/Borg information.
* **Tests**
  * Added a unit test covering agent-managed `ssh://...` repository creation and verifying the saved execution/executor settings.

### DIFF
--- a/app/api/repositories.py
+++ b/app/api/repositories.py
@@ -2006,11 +2006,16 @@ def _reject_agent_repository_ssh_target(
     connection_id: Optional[int],
     execution_target: Optional[str],
 ) -> None:
-    if (
-        connection_id
-        or (path or "").strip().startswith("ssh://")
-        or (execution_target or "").strip().lower() == "ssh"
-    ):
+    """Reject only a server-managed SSH connection for an agent repository.
+
+    An agent reaches its repository over its OWN SSH setup (mounted key /
+    known_hosts), so ``ssh://`` paths and ``execution_target="ssh"`` are fully
+    supported: the agent backup/operation payload already carries the path +
+    remote_path and the agent runs ``borg ... --remote-path`` itself. Only a
+    server-managed ``connection_id`` (an ``ssh_connections`` row that only the
+    Borg UI server can use, not the agent) stays unsupported.
+    """
+    if connection_id:
         raise HTTPException(
             status_code=400,
             detail={"key": "backend.errors.repo.agentRepositorySshUnsupported"},
--- a/tests/unit/test_api_repositories.py
+++ b/tests/unit/test_api_repositories.py
@@ -904,6 +904,60 @@ class TestRepositoriesCreate:
         assert repo.connection_id is None
         assert repo.repository_type == "local"
 
+    def test_create_agent_repository_accepts_ssh_path(
+        self, test_client: TestClient, admin_headers, test_db
+    ):
+        """Agent repos may target ssh:// with an explicit ``execution_target=ssh``;
+        the agent uses its own SSH credentials, so the server accepts it."""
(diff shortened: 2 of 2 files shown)


karanhudia:
[codex] Initial triage from BOR-185: I verified the current head `4e6d07cc` locally. `CodeRabbit` is passing; the current test already covers `execution_target="ssh"` with an agent-owned `ssh://` repository, so the earlier nitpick looks stale. Local checks run: `.venv/bin/python -m pytest tests/unit/test_api_repositories.py` (136 passed), `.venv/bin/ruff check app tests`, and `.venv/bin/ruff format --check app tests`. This still needs maintainer review/merge decision.

karanhudia:
This needs a UI follow-up.

When a Managed Agent is selected, users should be able to choose an agent-accessible remote repository target through the UI, similar to how Borg UI exposes Remote Machines for server-managed SSH repositories.

That likely means managed agents need to report the remote repository targets / SSH destinations they have access to, and the repository wizard should let users select one instead of requiring a manually typed `ssh://...` URL.

Otherwise the backend capability in this PR is not reachable through the intended product flow.

I will merge this so that you are not blocked, but I would appreciate a UI follow-up.

ioanalytica:
Thanks, Karan — and thanks for merging both PRs. You are right that the backend capability is not reachable through the intended product flow on its own. I will take the UI follow-up on as soon as I can: managed agents advertising the remote/SSH targets they can reach, and the repository wizard letting users pick one instead of hand-typing an `ssh://…` URL. I will open it as a separate PR and would be glad to sync on the approach (how agents should report their reachable targets) before going deep. Thanks again for the quick and thoughtful review.

PR #721, opened 2026-07-03
feat(repositories): UI support for agent-owned ssh:// repositories (follow-up to #708)

ioanalytica:
## Context

Follow-up to #708. That PR let the backend accept `ssh://` repositories owned by
a managed agent; this PR adds the UI and the agent-side plumbing so the flow
works end-to-end. It also opens a design question I'd rather think through
**together with you** than settle unilaterally — see *Architecture* below.

## The idea

There are hosts that are already prepared — by other means (Ansible, Kubernetes
pods, …) — to run Borg backups:

* they **know** the repository assigned to them,
* they hold SSH keys registered at the repository target,
* they have a `BORG_PASSPHRASE` ready.

In short: such a host can already operate its own repository with no further
configuration. That makes it an **ideal managed agent** — it can simply *tell*
borg-ui what the UI needs to know, instead of the UI having to provision SSH
connections and secrets for it.

The one thing deliberately **not** sent to the server is the secret
(`BORG_PASSPHRASE`) — the server doesn't need it, so it never learns it.

## What this PR does

* **The agent advertises its own repository.** New WS command
  `agent.repository_defaults` (+ endpoint
  `GET /managed-machines/agents/{id}/repository-defaults`): the agent reports its
  `$BORG_REPO` / `$BORG_REMOTE_PATH`, and whether a `$BORG_PASSPHRASE` is set —
  as a **boolean, never the value**.
* **The wizard auto-fills from that.** For an agent-owned repo the wizard keeps
  the full `ssh://` URL (instead of splitting it into a server SSH connection +
  path) and pre-fills repo path / remote-path from what the agent advertised; it
  warns when the agent reports nothing of its own.
* **Passphrase optional, with a server-side safeguard.** Since the agent runs
  every Borg operation itself, the server never needs the passphrase. The field
  becomes optional for agent repos — but only if the agent confirmed (via the
  boolean above) that it *has* one; otherwise create is refused, so you can't end
  up with an encrypted repo nobody can open.
* **Guards narrowed.** The remaining "agent repos can't use `ssh://` targets"
  guards now reject only a *server-managed* `connection_id` (repository create +
  backup-plan routing, front and back end).

**Security invariant:** every Borg operation for an agent-owned repo runs on the
agent with its own `$BORG_PASSPHRASE`. No secret ever reaches the server.

## Architecture — worth discussing together

This works and I'm running it, but it introduces a *model*: **a managed agent
that owns, and is self-sufficient about, its repository.** That's a slightly
different mental model from the current "server provisions everything" one, and I'd
rather align on it with you than push a shape you may want different. Points I'd
value your take on:

* Is "the agent advertises its repo defaults on demand" the right mechanism, or
  would you prefer the agent to register its repository explicitly at enrolment?
* Is a `has_passphrase` **boolean** the right amount for the server to know about
  the secret, or would you model that trust boundary differently?
* Browse / download / restore for a repo whose passphrase the server never holds
  all dispatch to the agent today — is that how you'd want it surfaced in the UI?

Treat this as a proposal to refine, not a fait accompli — happy to iterate.

## Notes

* Restore for agent-owned repos still has a permissions issue that's being scoped
  separately; it's not part of this PR. Backup, listing and browse work.

## Summary by CodeRabbit

* **New Features**
  * Added an admin-only endpoint for managed agents to report repository defaults (repo, remote path, and passphrase availability).
  * Repository wizard now fetches and uses these defaults for agent-managed setups.

* **Bug Fixes**
  * Encrypted repository checks now rely on whether the agent reports passphrase availability.
  * Improved handling of managed-agent SSH repository URLs in route planning/preview and wizard edit mode.
  * Passphrase validation rules were refined for agent vs non-agent targets; added user warnings when agent repo info is unknown.

* **Tests**
  * Updated and extended repository wizard tests for repository-defaults retrieval and URL preservation.

* **Documentation**
  * Added new user-facing translations for the new warning/error messages.

### DIFF
--- a/agent/borg_ui_agent/session.py
+++ b/agent/borg_ui_agent/session.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import json
 import logging
+import os
 import socket as socket_module
 import threading
 import time
@@ -333,6 +334,10 @@ class AgentSessionRuntime:
             self._handle_diagnostics(client, payload)
             return
 
+        if command == "agent.repository_defaults":
+            self._handle_repository_defaults(client, payload)
+            return
+
         if command == "cancel":
             # Signal the worker running this job so it actually stops; the worker
             # emits the job_canceled frame itself as it unwinds.
@@ -405,6 +410,21 @@ class AgentSessionRuntime:
         with self._registry_lock:
             self._cancel_events.pop(job_id, None)
             self._pending_cancels.discard(job_id)
+
+    def _handle_repository_defaults(
+        self, client: SessionCommandClient, payload: dict[str, Any]
+    ) -> None:
+        """Report the agent's own environment-configured repository target
+        (``$BORG_REPO`` / ``$BORG_REMOTE_PATH``) so the UI can pre-fill the
+        repository form, plus whether a ``$BORG_PASSPHRASE`` is set — a boolean,
+        never the passphrase value itself."""
+        client.send_result(
+            {
+                "repo": os.environ.get("BORG_REPO"),
+                "remote_path": os.environ.get("BORG_REMOTE_PATH"),
+                "has_passphrase": bool(os.environ.get("BORG_PASSPHRASE")),
+            }
+        )
 
     def _handle_filesystem_browse(
         self, client: SessionCommandClient, payload: dict[str, Any]
--- a/app/api/managed_machines.py
+++ b/app/api/managed_machines.py
@@ -544,6 +544,48 @@ async def browse_agent_machine_filesystem(
     )
 
 
+@router.get("/agents/{agent_machine_id}/repository-defaults")
+async def get_agent_machine_repository_defaults(
+    agent_machine_id: int,
+    _: User = Depends(require_managed_agents_admin_user),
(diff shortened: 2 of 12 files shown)


ioanalytica:
Re the nitpick suggesting Storybook coverage for the agent-repo warning banner (`RepositoryWizard.tsx` ~1069–1123): skipping, with a reason.

The banner is rendered by the parent `RepositoryWizard` (right before `<WizardStepLocation>`), gated on `agentRepoAdvertised` — which is `RepositoryWizard`'s own state, not a prop of `WizardStepLocation`. So a `WizardStepLocation` story structurally can't exercise it, and there's no `RepositoryWizard` story today (it's a full dialog with substantial API/query/analytics wiring). Adding Storybook coverage for this banner is out of scope for this PR.

karanhudia (approved):
Approving. CI is green, the PR is mergeable against current main, and there are no unresolved review threads. Remaining notes are non-blocking/out of scope; the agent-owned ssh:// repository flow looks acceptable to merge.

PR #723, opened 2026-07-05
fix(auth): don't 500 on password login for users without a local password

ioanalytica:
`verify_password` raised `ValueError: Invalid salt` when a user's stored `password_hash` is missing or not a valid bcrypt hash — e.g. OIDC-only users have no local password — surfacing as a **500** on `POST /api/auth/login` instead of a clean 401. This returns False for a missing/invalid hash so the login fails normally. Adds unit coverage for empty/None/invalid hashes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved password verification so missing, empty, or invalid stored hashes now safely fail instead of causing login errors.
  * Prevented invalid password hash formats from surfacing as server-side failures during sign-in.

* **Tests**
  * Added coverage for missing, empty, malformed, and valid password hash cases to confirm login behaves correctly.

### DIFF
--- a/app/core/security.py
+++ b/app/core/security.py
@@ -43,11 +43,22 @@ LOGIN_CHALLENGE_EXPIRE_MINUTES = 10
 TOTP_SETUP_EXPIRE_MINUTES = 15
 
 
-def verify_password(plain_password: str, hashed_password: str) -> bool:
-    """Verify a password against its hash"""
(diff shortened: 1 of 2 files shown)


karanhudia (approved):
Reviewed: narrow auth bug fix for invalid/missing password hashes; CI is green and mergeability is clean.

PR #724, opened 2026-07-05
fix(oidc): accept dict-shaped group/role claims (e.g. Zitadel)

ioanalytica:
`normalize_groups_claim` only handled list/string claims; a **dict** claim fell through to `[]`. Zitadel emits project roles as `urn:zitadel:iam:org:project:roles: {"admin": {...}, ...}` (role names as keys), so the admin allow-list never matched and OIDC admins were mapped as viewers. This treats a dict claim's keys as the group names (list/string paths unchanged). Adds unit tests incl. the Zitadel shape.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved support for OIDC group and role claims coming in as dictionary-shaped data, so group names are now recognized correctly.
  * Fixed claim normalization so empty, blank, and unsupported values are ignored consistently, while valid entries are preserved.
  * Added coverage to ensure common claim formats are handled reliably.

### DIFF
--- a/app/core/oidc.py
+++ b/app/core/oidc.py
@@ -389,6 +389,16 @@ def normalize_role_claim(value: Any) -> Optional[str]:
 
 
 def normalize_groups_claim(value: Any) -> list[str]:
+    if isinstance(value, dict):
+        # Some IdPs express roles/groups as an object keyed by the group name
+        # rather than a list — notably Zitadel's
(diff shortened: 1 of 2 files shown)


karanhudia (approved):
Reviewed: narrow OIDC claim normalization fix for dict-shaped group claims; CI is green and mergeability is clean.

PR #729, opened 2026-07-06
Make the managed-agent line functional (archive naming, browse/download, orphan recovery)

ioanalytica:
The managed-agent path (agents running borg on remote nodes, reporting back over
the WebSocket control channel) had a chain of bugs that made browse, download and
failure-recovery effectively non-functional. This PR fixes that line end to end.

The fixes are interdependent and build on each other, so the individual commits
are kept — please **review commit-by-commit**; each is a self-contained bug with
its root cause and fix in the commit body.

## Fixes, in order

1. **capture resolved archive name from `borg create --json`** — agents stored the
   unresolved `{now:…}` name template instead of the archive borg actually created,
   so browse/restore/delete/prune could never match the real archive.
2. **surface archive file-download errors as a toast instead of raw JSON** — the
   download navigated the browser to the URL, so any backend error rendered as a
   raw JSON page; fetch as a Blob and surface errors properly.
3. **reap orphaned in-flight agent jobs on a timer** — a job whose agent session
   died mid-run stayed `running` forever and wedged the repository via admission
   control; a background reaper terminally fails such jobs.
4. **fail orphaned request-scoped repository jobs instead of restarting them** —
   extract/list/info have no receiver once their HTTP request has timed out;
   restarting them on reconnect re-ran work for nobody and re-broke the agent
   session. Fail them terminally; keep retry-on-reconnect only for durable jobs
   (backups, maintenance, rclone_sync).
5. **stream archive file downloads over HTTP instead of base64-over-WS** — extract
   returned the whole file base64-encoded in a single WebSocket message, breaking
   the session for anything past the WS message limit (and taking concurrent jobs
   down with it). New agents advertise an `artifact.upload` capability and stream
   `borg extract --stdout` straight to a new artifact endpoint, proxied to the
   client via a StreamingResponse (any size, constant memory, no disk buffer),
   with a download activity/progress toast and busy-disable in the UI.

## Backward compatibility

All server changes remain compatible with older agents: the streaming download is
gated on an agent capability and falls back to the previous base64 path. Nothing
here breaks agents that predate this PR.

## Testing

Unit tests added (artifact relay, job reaper, request-scoped classification, agent
streaming upload, download UI). Verified end-to-end on a live managed-agent
deployment (Hetzner Storage Box over ssh, borg1): browse, small and large (85 MB
etcd snapshot) downloads with live progress, and automatic orphaned-job recovery.

## Summary by CodeRabbit

* **New Features**
  * Added streaming artifact uploads/downloads for capable agents, including an HTTP artifact relay and a streaming archive download path with first-byte/idle timeouts.
  * Introduced an agent job reaper to automatically fail stale in-flight jobs (avoids requeueing for request-scoped operations).
  * Extended archive download UI for richer preparation/progress/completion states and size-aware progress.

* **Bug Fixes**
  * Improved backup completion reporting to show the resolved archive name when available.
  * Hardened artifact transfer handling for missing consumers and cancellation/timeout scenarios.

* **Maintenance**
  * Updated package version to 0.1.1.


ioanalytica:
@coderabbitai review

karanhudia (approved):
Approving. CI is green, the PR is mergeable against current main, and there are no unresolved review threads. This is a larger managed-agent behavior change, so after merge/deploy we should monitor archive downloads, artifact streaming, and stale-job reaping.

PR #730, opened 2026-07-06
feat(restore): delegate managed-agent restores and restore-checks to the agent

ioanalytica:
Both ad-hoc restores and scheduled restore-checks ran `borg extract` **server-side**, which cannot work for agent-only repositories: the server has no access to the node's filesystem (and, for agent-owned repos, often not to the repository either). This PR makes the restore path agent-aware, mirroring how backups are already delegated to the agent.

## What it does

- Adds a shared `repository.restore` agent job kind. The agent runs `borg extract` **on the node**, with `cwd` set to the resolved restore target, streams progress back over the control channel, and — for restore-checks — verifies the canary on the node.
- `restore_service` and `restore_check_service` detect `executor_type=agent` repositories and delegate through the new job kind instead of extracting locally. Non-agent repositories keep the existing server-side path unchanged.
- Restore-target resolution supports a throwaway temp dir (used by restore-checks, cleaned up afterwards) and a caller-supplied destination on the node.

## Components

- `agent/borg_ui_agent/repository_ops.py` — `repository.restore` execution: extract on the node, progress streaming, cancellation, canary verification.
- `agent/borg_ui_agent/runtime.py` — register the `repository.restore` capability + handler.
- `app/services/restore_service.py` — `_is_agent_restore` / `_execute_agent_restore` delegation + await of the agent job.
- `app/services/restore_check_service.py` — agent branch for scheduled restore-checks.
- `app/services/repository_executor.py`, `app/services/job_admission.py` — wire the new job kind through the executor and admission control.

## Testing

- Unit: `tests/unit/test_agent_restore.py`, `tests/unit/test_restore_agent_delegation.py`, `tests/unit/test_restore_service.py` — **42 passing**, `ruff` clean.
- Validated end-to-end on a real Kubernetes node: an agent-only repository restore extracted the selected file on the node (writable destination) via the delegated job.

## Note

The restore extracts into the destination on the node, so that path must be writable by the agent. Restoring into a read-only mount makes borg skip the file with a warning — a deployment concern, not a restore-path issue.

## Summary by CodeRabbit

* **New Features**
  * Added managed repository restore jobs with agent capability discovery, progress mirroring, canary-based verification, and user cancellation support.
  * Enabled restore job support across admission, executor capability validation, and repository restore checks (including delegated agent-executor flow).

* **Bug Fixes**
  * Improved restore/verification handling by treating Borg warning exit codes as non-fatal and enforcing safe, within-root canary/path validation.
  * Ensured temporary restore destinations are cleaned up after completion.

* **Tests**
  * Added unit coverage for restore command generation (Borg 1/2), agent delegation/routing, canary verification outcomes, and timeout/notification behavior.


ioanalytica:
@coderabbitai review

Squashed to a single commit; the four earlier findings are addressed (borg2 `--progress`, `--` before restore paths, bounded agent-restore/​restore-check waits, and mirrored restore notifications).

ioanalytica:
@coderabbitai review

Addressed the remaining nitpick in `bedc51de`: the agent-restore wait test now derives its fake clock from `_AGENT_RESTORE_CLAIM_TIMEOUT_SECONDS` instead of hardcoded magic numbers, so it stays correct if the timeout constant changes. Amended into the single squashed commit.

karanhudia (approved):
Approving. CI is green, the PR is mergeable against current main, and the previously raised restore-delegation concerns have been addressed in later commits. Since this enables managed-agent restore execution, we should monitor the first restore and restore-check runs closely after deploy.

PR #734, opened 2026-07-07
fix(agent): keep idle agents live via a WS heartbeat message

ioanalytica:
## Problem

Managed agents keep their WebSocket alive with **protocol-level** pings, which the server library auto-pongs. But those pings never reach application code, so the server's `last_seen_at` only advances when the agent is actively `POST /heartbeat`-ing during a job (the 5s `should_cancel` check).

An **idle-but-healthy** agent therefore freezes at "time of last job" and looks stale — even though its WS is connected and fine. On a busy cluster, every agent's `last_seen_at` sat hours in the past while all were online.

## Fix

- **Agent** (`session.py`): the idle recv-timeout branch — which fires every ~30s and keeps running even while a job runs in a worker thread — now also sends an application-level `{"type": "heartbeat"}` frame, alongside the existing protocol ping.
- **Server** (`agents.py` `_handle_agent_session_message`): handles `heartbeat` by refreshing `last_seen_at` and marking the agent `online`.

`last_seen_at` now reflects real "last contact" instead of "last job", which is a prerequisite for liveness-based orphaned-job handling.

## Tests

- `test_agent_heartbeat.py`: the session handler refreshes `last_seen_at` / status on a `heartbeat` message.
- `test_agent_runtime.py`: the session emits an app heartbeat on an idle recv timeout (existing idle-ping test updated for the new frame).

## Summary by CodeRabbit

* **Bug Fixes**
  * Agent sessions now send a lightweight heartbeat during idle periods, helping keep connections alive more reliably.
  * Heartbeat messages are now recognized by the backend and update agent status and last-seen time correctly.
  * Improved idle-session handling so agents stay marked online instead of being treated as inactive.

### DIFF
--- a/agent/borg_ui_agent/session.py
+++ b/agent/borg_ui_agent/session.py
@@ -233,11 +233,20 @@ class AgentSessionRuntime:
                 try:
                     raw_message = socket.recv()
                 except (TimeoutError, WebSocketTimeoutException):
-                    # Idle keep-alive ping. The key property of this loop is that
-                    # it stays in recv() *even while a job runs in a worker
-                    # thread*, so websocket-client keeps auto-ponging the server's
-                    # pings and the session survives long backups/checks.
+                    # Idle keep-alive. The key property of this loop is that it
+                    # stays in recv() *even while a job runs in a worker thread*,
+                    # so websocket-client keeps auto-ponging the server's pings
+                    # and the session survives long backups/checks.
+                    #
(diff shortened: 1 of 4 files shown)


karanhudia (approved):
Approving. CI is green, the PR is mergeable against current main, there are no unresolved review threads, and the app-level heartbeat is a small targeted fix for idle managed-agent liveness. Good to merge before the larger agent session changes.

PR #735, opened 2026-07-07
fix(agent): deliver terminal job outcomes over REST, not the WebSocket

ioanalytica:
## Problem

A managed agent runs each session command in a **worker thread** and reported the outcome (`command_result` / `command_error`) back over the **same WebSocket** the recv loop uses for keepalive.

A job result can be large — an archive listing is hundreds of KB. Sending it as one WebSocket frame holds `websocket-client`'s per-connection send lock for the whole write, which **starves the keepalive pong** the recv loop must emit. The server then hits its ws-ping timeout (uvicorn default **20 s**) and closes the session **mid-send** → the result is lost and the job stays in-flight until the reaper fails it **15 minutes** later.

Symptom: **archive browsing returns an empty dialog** (and the repo is briefly blocked for other operations by admission), while `borg list` on the node itself is fast.

```mermaid
sequenceDiagram
  participant Srv as Server (uvicorn)
  participant Ag as Agent recv-loop
  participant Wk as Worker thread
  Srv->>Ag: command (list_archive_contents)
  Note over Wk: runs borg list → ~660 KB result
  Srv-->>Ag: ws PING (every 20s)
  Wk->>Srv: command_result (660 KB) — holds WS send-lock
  Note over Ag: recv-loop can't PONG (send-lock held by worker)
  Srv--xSrv: ws-ping timeout (20s) → close connection
  Wk--xSrv: send fails: Broken pipe → result lost
  Note over Srv: job orphaned → reaped after 15 min → empty dialog
```

## Fix

Terminal outcomes for **persisted jobs** are delivered over the **REST job API** (`/jobs/{id}/complete` · `/fail` · `/cancel`) instead of the WebSocket. The socket then only carries small control frames + keepalive, so a large result can no longer starve the pong.

```mermaid
sequenceDiagram
  participant Srv as Server
  participant Ag as Agent recv-loop
  participant Wk as Worker thread
  Srv->>Ag: command (list_archive_contents)
  Note over Wk: runs borg list
  Srv-->>Ag: ws PING → Ag auto-PONGs (socket free)
  Wk->>Srv: POST /jobs/{id}/complete (result) — REST
  Srv-->>Wk: 200 OK (idempotent)
  Note over Srv,Ag: WS carries only control + keepalive → session stays healthy
```

Details:
- `/complete`, `/fail`, `/cancel` are **idempotent** (no-op on an already-final job), so a retry or a late delivery is harmless.
- The REST client wraps one `requests.Session` shared by worker threads, so terminal calls are **serialized under a lock**.
- **Ephemeral** commands (no job row — e.g. an interactive filesystem browse) still return their small result over the WS, correlated by `command_id`.
- Telemetry frames (`job_started` / `progress` / `log`) remain best-effort WS sends.

## Scope / risk

This only touches the **managed-agent WebSocket session** — a new code path that is not yet in service — so there is no impact on server-side execution or existing repositories. `command_result` handling on the server WS side is left intact (unused by persisted jobs now).

## Tests
- results go over REST, not the WS (persisted job)
- `command_error` → `fail_job` with `error_message`/`return_code` preserved (per review)
- ephemeral (job_id = None) result still returns over the WS
- idle heartbeat / durable job events unchanged otherwise

## Summary by CodeRabbit

* **New Features**
  * Added an optional REST fallback for terminal job outcomes (completion, failure, and cancellation) when live terminal delivery isn’t available.
  * When REST delivery is used, terminal outcome frames are sent via the job API instead of over the live terminal channel.

* **Bug Fixes**
  * Made live terminal delivery best-effort so transient send failures don’t disrupt processing.
  * REST delivery errors are handled without preventing session completion.
  * Cancellation outcomes now flow through the dedicated terminal outcome path.

* **Tests**
  * Added/updated unit tests to verify REST job API delivery and to ensure terminal frames are not emitted over the live channel when REST is used.

### DIFF
--- a/agent/borg_ui_agent/session.py
+++ b/agent/borg_ui_agent/session.py
@@ -69,6 +69,8 @@ class SessionCommandClient:
         send_lock: Optional[threading.Lock] = None,
         closing: Optional[threading.Event] = None,
         artifact_uploader: Optional[Callable[[int, Any], dict[str, Any]]] = None,
+        http_client: Optional[AgentClient] = None,
+        http_lock: Optional[threading.Lock] = None,
     ):
         self.socket = socket
         self.command_id = command_id
@@ -78,6 +80,12 @@ class SessionCommandClient:
         self._send_lock = send_lock or threading.Lock()
         self._closing = closing
         self._artifact_uploader = artifact_uploader
+        # Terminal job outcomes (result/error/canceled) are delivered over the
+        # REST job API, not the WebSocket — see _deliver_terminal. The client
+        # wraps one requests.Session shared by all worker threads, so calls are
+        # serialized under http_lock.
+        self._http_client = http_client
+        self._http_lock = http_lock or threading.Lock()
 
     def upload_artifact(self, job_id: int, data: Any) -> dict[str, Any]:
         """Stream a binary job artifact to the server over HTTP (not the WS)."""
@@ -118,12 +126,9 @@ class SessionCommandClient:
     def complete_job(self, job_id: int, *, result: dict[str, Any]) -> dict[str, Any]:
         self._ensure_started(job_id)
         self.finished = True
-        self._send(
-            {
-                "type": "command_result",
-                "job_id": job_id,
-                "result": result,
-            }
+        self._deliver_terminal(
+            {"type": "command_result", "job_id": job_id, "result": result},
+            lambda c: c.complete_job(job_id, result=result),
         )
         return {"id": job_id, "status": "completed"}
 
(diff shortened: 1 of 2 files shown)


karanhudia:
Thanks for this one. The direction still makes sense to me, but after landing the managed-agent base PRs this now conflicts with current main in `agent/borg_ui_agent/session.py`.

Could you rebase/update this branch and resolve that conflict? Once the final diff is clean and CI is still green, I will review it again for approval.

ioanalytica:
Rebased onto current main and resolved the `session.py` conflict. The overlap was orthogonal — the managed-agent base added the artifact-upload path (`artifact_uploader`/`_artifact_client`) while this PR adds the REST terminal-delivery path (`http_client`/`http_lock`); kept both in the constructor, `__init__` and the `SessionCommandClient` call. Single clean commit on top of main, agent runtime tests green (53 passed, incl. both the heartbeat and REST-delivery tests). Ready for another review pass.

PR #756, opened 2026-07-15
fix(agent): honor configured list/info timeout on agent archive+info endpoints

ioanalytica:
## Problem

For managed-agent repositories, opening the repository view shows **no archives at all**. The archive list (and the info/stats panel) come up empty even though the repository has archives.

## Root cause

`GET /{repo_id}/archives` (`list_repository_archives`) and `GET /{repo_id}/info` (`get_repository_info`) delegate to the managed agent when `is_agent_executor(repository)`, but they awaited the agent job with the **bare 15s default** of `wait_for_agent_repository_operation_job(db, agent_job.id)`:

```python
result = await wait_for_agent_repository_operation_job(db, agent_job.id)
```

A Borg 2 `repo-list` / `info` run on the agent against a slow remote (e.g. a Storage Box reached over plain/dumb sftp with no server-side borg) routinely takes ~20s. That exceeds the 15s default, so the endpoint raises `504 repositoryOperationTimeout` and the UI renders an empty archive list.

The sibling Borg 2 delegation call sites in the same file already pass the configurable timeout (`get_operation_timeouts(db)["list_timeout"|"info_timeout"]`, default 600s); these two endpoints were simply missed.

## Fix

Pass the configured `list_timeout` / `info_timeout` to both endpoints, matching the existing pattern:

```python
result = await wait_for_agent_repository_operation_job(
    db,
    agent_job.id,
    timeout_seconds=get_operation_timeouts(db)["list_timeout"],  # / info_timeout
)
```

The wait still returns as soon as the agent job completes or fails, so the happy path is unchanged — this only lifts the artificial 15s ceiling for slower remotes.

## Testing

- Reproduced against a managed agent backing a Borg 2 sftp repository: `GET /{repo_id}/archives` and `/info` consistently 504'd at 15s while the underlying `repository.list_archives` agent job completed in ~20s and returned valid JSON. With the fix the endpoints return the archive list.
- `python -m py_compile app/api/repositories.py` passes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of agent-backed repository archive listings and repository information requests by applying the appropriate operation timeouts.
* **Tests**
  * Updated unit tests to configure distinct timeouts for “info” vs “list/archives” operations and verify the correct timeout value is used when awaiting agent jobs.

### DIFF
--- a/app/api/repositories.py
+++ b/app/api/repositories.py
@@ -5958,7 +5958,11 @@ async def list_repository_archives(
                 job_kind="repository.list_archives",
             )
             await dispatch_agent_job_best_effort(db, agent_job, repository_id=repo_id)
-            result = await wait_for_agent_repository_operation_job(db, agent_job.id)
+            result = await wait_for_agent_repository_operation_job(
+                db,
+                agent_job.id,
+                timeout_seconds=get_operation_timeouts(db)["list_timeout"],
+            )
             archives_data = _parse_agent_json_result(result)
             archives = archives_data.get("archives", [])
             archives = enrich_archives_with_backup_metadata(archives, repository, db)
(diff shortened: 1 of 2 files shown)


karanhudia (approved):
Reviewed the timeout propagation and its endpoint-specific coverage; approved for merge.

PR #757, opened 2026-07-15
fix(agent): stop persisting browse listings in the DB and bound the job list

ioanalytica:
## Impact & urgency

Not a new regression — the persisted browse listing was introduced in `f075bce4` ("fix(agent): dispatch backups and browse archives via agent"), so it has shipped in **every release since v2.2.0**. On managed-agent installs `agent_jobs.result` grows unboundedly (one install reached ~2 GB within a week) and `GET /managed-machines/agent-jobs` then stalls loading it. Because it affects all `v2.2.x`, landing this in the next tagged release would be valuable; the included migration also reclaims the already-accumulated space on existing installs.

## Problem

Two issues around managed-agent archive browsing compound into unbounded app-DB growth and a stalled admin UI:

1. **Browse listings are persisted forever.** A completed `repository.list_archive_contents` job stores the archive's full `borg list` output in `agent_jobs.result`. For a large archive that is hundreds of MB, and it is never cleared — every browse adds another copy. The listing is purely transient: it only needs to survive until the polling browse request reads it once, after which the parsed items live in the archive cache and the agent can reproduce them on demand.

2. **`GET /managed-machines/agent-jobs` returns everything, including results.** The endpoint returns *all* agent jobs with their full `result` blobs and no limit. Once a few large browse results have accumulated, a single call loads and serializes gigabytes and blocks the event loop for tens of seconds — so the managed-agent page (which polls this endpoint) can no longer refresh, and every concurrent request waits behind it.

## Fix

- **browse**: clear `agent_jobs.result` once the listing has been consumed into the cache (guarded in `finally`; the 202 "still pending" path keeps it so the next poll can still read it).
- **agent-jobs list**: never load or serialize `result` — SQLAlchemy `defer(AgentJob.result)` plus a result-less `AgentJobSummaryResponse` — and bound the row count with a `limit` query param (default 200, max 1000). The list view only needs metadata; a single-job fetch still carries the result.
- **migration**: clear `result` for existing `repository.list_archive_contents` jobs and VACUUM to return the freed space to the filesystem (startup-only, best-effort).

## Tests

- `_drop_agent_job_result` nulls a persisted listing.
- The job list omits `result` and honours `limit`.
- Existing browse tests (queue/poll/202/resume) still pass.

## Summary by CodeRabbit

* **New Features**
  * Agent job listings now support a bounded `limit` (newest-first).
  * Agent job listings return metadata only, omitting the large result payload.
* **Bug Fixes**
  * Browse returns HTTP 202 with a job ID while agent results are still pending.
  * After browse results are consumed, stored results are cleared to reduce payload retention (including for previously persisted cached listings via an SQLite-safe migration).
* **Tests**
  * Added/expanded unit tests for result clearing, listing shape, newest-first ordering, and `limit` behavior.

### DIFF
--- a/app/api/browse.py
+++ b/app/api/browse.py
@@ -6,7 +6,7 @@ from sqlalchemy.orm import Session
 import os  # noqa: F401
 import structlog
 
-from app.database.models import User, Repository, SystemSettings
+from app.database.models import User, Repository, SystemSettings, AgentJob
 from app.database.database import get_db
 from app.api.auth import get_current_user
 from app.core.borg_router import BorgRouter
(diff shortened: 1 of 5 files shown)


ioanalytica:
_Moved into the PR description — see the **Impact & urgency** section at the top._

karanhudia (approved):
Reviewed the bounded metadata listing and browse-result cleanup; approved for merge.

PR #759, opened 2026-07-16
Fix timeline ordering and harden release validation

karanhudia:
## Description
Fixes the dashboard timeline ordering from #758 and hardens the release
validation changes already included on this branch. Same-day activity dots now
render oldest-to-newest, while release workflows safely validate inputs and
only publish main-ancestry revisions.

## Related Issue
Fixes #758

## Type of Change
- [x] 🐛 Bug fix (non-breaking change which fixes an issue)
- [ ] ✨ New feature (non-breaking change which adds functionality)
- [ ] 💥 Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] 📖 Documentation update
- [x] 🎨 UI/UX improvement
- [x] 🧪 Test addition/improvement
- [ ] ♻️ Refactoring

## Testing
- [x] I have tested this locally
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] All existing tests pass

Commands run:
- `bash -n scripts/semver.sh scripts/check-release-version.sh scripts/bump-version.sh scripts/release.sh`
- strict-SemVer parser checks for valid and leading-zero versions
- `npm test -- --run src/pages/dashboard-v3/ActivityTimeline.test.tsx`
- `npm run check:locales`
- `npm run typecheck`
- `npm run lint`
- `npm run build`

## Checklist
- [x] I have read the [CONTRIBUTING](CONTRIBUTING.md) guidelines
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] New and existing unit tests pass locally with my changes

## Screenshots (if applicable)
Not applicable; this corrects the ordering of existing timeline dots.

## Additional Context
Release workflow values now pass through environment variables, manual
dispatches verify main ancestry, SemVer rules are shared by all release
commands, and main plus its release tag are pushed atomically.

## Summary by CodeRabbit

* **New Features**
  * Added stricter release validation to ensure the release tag, version metadata, and source commit ancestry are consistent before publishing.
  * Added gating so Docker image publishing only proceeds after release validation completes.
  * Added Storybook coverage for the activity timeline.
* **Bug Fixes**
  * Activity timeline events now render in deterministic chronological order within the same day.
* **Documentation**
  * Documented the release process, supported versioning, verification behavior, and guidance for immutable published tags.
* **Tests**
[...]

PR #761, opened 2026-07-17
Hardening of the hardening of the release validation

ioanalytica:
The release validation added in #759 is a good idea — and it currently fails against `main` itself, which blocks the next release:

```
$ ./scripts/check-release-version.sh v2.2.6
Error: app/config.py declares 2.2.3; expected 2.2.6 for v2.2.6.
```

`VERSION` and `frontend/package.json` say `2.2.6`; `app/config.py` and `app/main.py` still say `2.2.3`. The 2.2.6 bump (`619db52d`) did not go through `release.sh`, which is the only thing that rewrites those two literals, so they kept what they last said — and the check that would have caught it landed afterwards. It is wired into `release.yml`, `docker-publish.yml` and `release.sh`, so the next tag hits it.

## Why not just bump the two literals

That fixes today and leaves the same trap set: they are copies, and a copy drifts again the moment a bump takes any other path. This removes them instead, so **`VERSION` is the only place a version is written down**:

- `app/config.py` reads `VERSION`
- `app/main.py` takes its version from `get_runtime_app_version()`
- nothing else declares one

**No version is bumped here.** `VERSION` stays at 2.2.6 — which release comes next is entirely your call. The only user-visible change is that `/api/docs` stops advertising 2.2.3: at runtime the app already reported 2.2.6, because `get_runtime_app_version()` reads the `VERSION` file.

## Two things found on the way

**`get_runtime_app_version()` looked for an absolute `/app/VERSION`**, so it only worked inside the image; everywhere else it silently fell through to the stale default. It now resolves `VERSION` relative to the package. The file still wins over `APP_VERSION`, deliberately: the image defaults that ARG to `dev`, so reading the environment first would make any image built without `--build-arg` report itself as `dev`.

**`app_version` was a `BaseSettings` field**, so pydantic refilled it from `APP_VERSION`. With `APP_VERSION=dev` — i.e. in that same image — `settings.app_version` was `"dev"` while `get_runtime_app_version()` said `2.2.6`. The field is removed rather than shielded from the environment; nothing read it, `app/api/settings.py` already calls `get_runtime_app_version()`.

## Tooling

`release.sh` no longer rewrites the literals and `check-release-version.sh` no longer greps for them — both would fail on code that no longer contains them. `release.sh` now writes two places instead of four, and its regex-rewriting Python block is gone.

`tests/unit/test_version.py` takes over that job and covers more than the regexes could: it fails if either literal is reintroduced, and pins the fallback chain (unreadable `VERSION` → `""` rather than an exception, then `APP_VERSION`, then `0.0.0`). That branch matters now that `VERSION` is the only source — `main.py` asks for the version while being imported.

## Frontend

It never had a version of its own. `__APP_VERSION__` was injected into the bundle from `package.json` by both `vite.config.ts` and `vitest.config.ts` and read by nothing — the sidebar shows `app_version` from the API. The dead constant is removed, which leaves `package.json`’s version as npm metadata (the package is `private`), unable to disagree with anything. `release.sh` still keeps it in step via `npm version`, and `check-release-version.sh` still enforces it.

## Verification

- `./scripts/check-release-version.sh v2.2.6` passes (was the error above)
- Setting `VERSION` alone to `2.2.7` makes the app report `2.2.7` with no code change; the check then correctly asks only for the frontend bump
- Reintroducing either literal fails `tests/unit/test_version.py` (checked both)
- Backend suite: 2560 passed. The 2 failures + 15 errors also occur on an untouched `main` checkout and are unrelated to this change — happy to open a separate issue if useful
- Frontend: `format:check`, `typecheck`, `lint`, `check:locales`, `build`, and 2302 vitest tests pass

## Summary by CodeRabbit

* **New Features**
  * Application and API documentation versions are now derived from the packaged release `VERSION`.
  * Added fallback version resolution using `APP_VERSION`, defaulting to `0.0.0` when unavailable.
* **Bug Fixes**
  * Improved robustness of version detection when packaged metadata is missing or unreadable.
  * Removed reliance on hardcoded version values during releases.
* **Tests**
  * Added/updated coverage for version consistency across API settings and published API metadata, including precedence and fallback behavior.


ioanalytica:
@karanhudia : Long story short: With this PR you would only need to update VERSION when bumping the version. You should also create the version tag. But even if this tag would ever be orphaned the app would report the correct version everywhere. Tests make sure that no inadvertent mistakes can happen.

karanhudia (approved):
Reviewed the single-source runtime version path, release-script updates, and fallback coverage; approved for merge.

PR #762, opened 2026-07-18
fix(reaper): reap orphaned pending maintenance jobs that block a repository

ioanalytica:
## Problem

A repository can be blocked forever by an orphaned maintenance job. The plan/maintenance flow creates a `PruneJob`/`CompactJob`/`CheckJob`/`DeleteArchiveJob` row and *then* queues the agent job for it. Two ways this leaves a permanently-`pending` row that no agent job will ever update:

1. **Queue failure after commit** — if queuing raises (e.g. `database is locked` under a concurrent multi-repo plan run), the `*_job` row is already committed but the agent job never gets created.
2. **Pre-existing stuck rows** — a crash between creating the row and queuing, or rows already stuck from an earlier version.

Admission control counts a `pending` maintenance job as active write work, so the repository stays blocked indefinitely. Verified live: `prune_jobs` stuck at 129 on one install.

## Change

Two layers — remove the root cause, then add a safety net.

**Fail-closed dispatch** (`_run_agent_maintenance`): wrap the queue call so that on *any* failure the linked `*_job` is marked `failed` (`_fail_orphaned_maintenance_job`) before the error propagates. The row can no longer be left `pending` with no agent job.

**Reaper reconciliation** (`reconcile_orphaned_maintenance_jobs`, run as a third pass in the reaper loop): fail a `*_job` that is `pending`, older than the existing 5-minute age guard, and has no active agent job. The agent job is correlated via the agent payload's `maintenance_job {kind, id}` (repository agent jobs carry no `backup_job_id`). Only `pending` is reaped — `running` rows are already covered by the existing process-liveness reapers, and server-side maintenance runs in-process without an agent job at all.

## Tests

- `tests/unit/test_borg_router.py` — dispatch failure marks the maintenance job failed instead of leaking a `pending` row.
- `tests/unit/test_utils.py` — reaper reconciles an orphaned `pending` maintenance job past the age guard and leaves `running`/correlated ones untouched.

Backend-only; no frontend or migration changes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented maintenance tasks from getting stuck as pending when agent queuing fails.
  * Added automatic reconciliation to fail orphaned, aged maintenance tasks that no longer have active agent work.
  * Preserved maintenance tasks that are still associated with active/queued agent operations.
* **Reliability**
  * Enhanced background reaping to include orphaned maintenance cleanup in each cycle.
  * Improved failed-job reporting with error details and completion timestamps when available.
* **Tests**
  * Added unit and database-backed coverage for queue failures, orphaned tasks, active associations, race handling, and age-threshold skipping.

### DIFF
--- a/app/core/borg_router.py
+++ b/app/core/borg_router.py
@@ -17,6 +17,48 @@ from typing import List, Optional
 logger = structlog.get_logger()
 
 
+def _fail_orphaned_maintenance_job(
+    db: Session, maintenance_kind: str, maintenance_job_id: int
+) -> None:
+    """Mark a maintenance ``*_job`` failed when its agent job could not be queued.
+
+    Without this the row stays 'pending' with no backing work and blocks the
+    repository via admission control until a reaper eventually clears it.
+    """
+    from datetime import datetime
+
+    from app.database.models import CheckJob, CompactJob, DeleteArchiveJob, PruneJob
+
+    models = {
+        "check": CheckJob,
+        "compact": CompactJob,
+        "prune": PruneJob,
+        "delete_archive": DeleteArchiveJob,
+    }
+    model = models.get(maintenance_kind)
+    if model is None:
+        return
+    try:
+        # The failed queue attempt (e.g. "database is locked") may have left this
+        # session's transaction unusable, which would make the query below raise
+        # and skip the update. Reset it first; the row was committed by the caller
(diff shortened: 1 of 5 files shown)


karanhudia (approved):
Reviewed the fail-closed maintenance path, guarded reaper transition, and concurrency coverage; approved for merge.

PR #764, opened 2026-07-18
fix(backups): preserve symlinks faithfully on SSH-source backups and restores

### LINKED ISSUE #751: SSH-source backups via sshfs silently lose symlinks and hard links (restore ≠ backup)
# SSH-source backups via sshfs silently lose symlinks and hard links (restore ≠ backup)

## Summary

When a backup **source** is a remote host reached over SSH, borg-ui mounts it with `sshfs` and runs `borg create` over the mountpoint. The mount is built by `build_sshfs_command()` in `app/services/mount_service.py`, which hardcodes `-o follow_symlinks`. As a result, **symbolic links and hard links are silently altered or dropped before borg ever sees them**, so a restore does not reproduce the source faithfully.

This is a data-fidelity bug, not just a preference: `borg create`/`borg check` cannot detect it, because from borg's point of view it faithfully stored exactly what the (sshfs) filesystem presented. The divergence only surfaces if someone compares a restore against the live source — which almost never happens.

Two independent loss classes, both reproduced empirically (details below):

1. **Symlinks** — dereferenced, and (surprisingly) even with `follow_symlinks` *off*, absolute symlinks are silently dropped due to sshfs's `contain_symlinks` sandbox. **Fixable** with the right mount options.
2. **Hard links** — always broken into independent copies. **Not fixable** at the sshfs layer (SFTP does not carry inode/nlink).

## Affected code path

Only **server-side SSH-source backups** are affected (source mounted via sshfs). Local-source and node-local (agent) backups read a real filesystem and are unaffected.

- Mount builder: `MountService.build_sshfs_command()` — `app/services/mount_service.py:1606`, hardcoded `-o follow_symlinks` at `:1628`.
- Backup source mount: `_prepare_source_paths()` → `mount_service.mount_ssh_paths_shared()` (`app/services/backup_service.py:1321`) → `_execute_sshfs_mount()` (`mount_service.py:1535`).
- The same builder is shared by browse/cloud-mirror (`mount_ssh_directory()`, `mount_service.py:356`, used by `rclone_repository_service.py:290`) and restore-to-SSH (`restore_service.py:1258`).

## Environment

- sshfs 3.7.3, FUSE 3.17.2
- borg 1.4.4
- remote: Linux (kernel 6.8), OpenSSH `Subsystem sftp` (SFTP protocol v3)

## Evidence 1 — symbolic links

Test tree on the remote source:

| name | type |
|---|---|
| `realfile.txt` | regular file |
| `realdir/inside.txt` | regular file in a dir |
| `link-to-file` → `realfile.txt` | relative symlink |
| `link-to-dir` → `realdir` | relative symlink to dir |
| `abs-link` → `/etc/hostname` | absolute symlink |
| `broken-link` → `/nonexistent/target` | broken absolute symlink |

What borg's `lstat` sees on the mount, per option set:

| mount options | relative symlinks | absolute symlinks | broken symlink | faithful? |
|---|---|---|---|---|
| **`follow_symlinks`** (current default) | → **dereferenced file / dir** (dir symlink is recursed → target tree **duplicated**) | → **dereferenced file content** | → **dropped** (unreadable) | ❌ no symlink survives |
| nofollow (sshfs default, `contain_symlinks` on) | stored as symlink ✅ | **`readlink` → EPERM → borg logs `E` → dropped** | **dropped** | ❌ absolute links silently lost |
| `transform_symlinks` (`contain_symlinks` on) | ✅ | **dropped** (contain still blocks) | **dropped** | ❌ same |
| **`no_contain_symlinks`** (nofollow) | ✅ | **`abs-link -> /etc/hostname` ✅** | **`broken-link -> /nonexistent/target` ✅** | ✅ fully faithful |

Observed with the current default (`follow_symlinks`), `borg extract` into an empty dir produced a **regular file** where each symlink was, `link-to-dir/` as a real directory (its contents duplicated with `realdir/`), and `broken-link` missing entirely.

Observed with nofollow: `readlink abs-link` → `Operation not permitted`; `borg create` logged `E abs-link` / `E broken-link` and omitted both from the archive.

Observed with `no_contain_symlinks`: all four symlinks stored with their original targets, incl. the absolute and broken ones.

Root cause: `contain_symlinks` (enabled by default in sshfs 3.7.3) rejects any symlink whose target escapes the mount root, which is why absolute links yield `readlink` EPERM and are dropped. This is safe to disable for backup because borg in nofollow mode only reads the target **string** (`readlink`); it never traverses it, so there is no sandbox escape.

## Evidence 2 — hard links (unfixable over sshfs)

Two hard links to the same inode on the source, traced end to end:

| stage | `f1` | `f2` |
|---|---|---|
| **source (remote)** | `nlink=2  inode=288810` | `nlink=2  inode=288810` (same inode → real hard link) |
| **over sshfs** | `nlink=1  inode=2` | `nlink=1  inode=3` (nlink reset, synthetic per-path inodes) |
| **`borg create --list`** | `A f1` | `A f2` (two independent files) |
| **after `borg extract`** | `nlink=1  inode=529412` | `nlink=1  inode=529411` (hard link lost) |

Control — the identical test on a **local** filesystem (no sshfs):

| stage | `g1` | `g2` |
|---|---|---|
| local source | `nlink=2  inode=529422` | `nlink=2  inode=529422` |
| after `borg extract` | `nlink=2  inode=529434` | `nlink=2  inode=529434` (hard link **preserved**) |

The control proves borg preserves hard links correctly; the loss happens purely at the sshfs/SFTP layer. **Root cause:** SFTP protocol v3 (OpenSSH) transmits neither `st_ino` nor `st_nlink`, so sshfs reports `nlink=1` and synthesizes a distinct inode per path. borg therefore never enters hard-link detection (which keys on `st_nlink > 1` + matching `(st_dev, st_ino)`). **No mount option can fix this** — the information is not on the wire.

Consequences: archive size is unaffected (chunk-level dedup still stores the content once), but the link structure is lost and a restore can consume far more disk than the source for hard-link-deduplicated trees (e.g. Nix store, rsnapshot rotations, `.git` object stores).

## Reproduction

On the remote source:
```sh
d=$(mktemp -d); echo x > "$d/f1"; ln "$d/f1" "$d/f2"          # hard links
ln -s realfile.txt "$d/link-rel"; ln -s /etc/hostname "$d/link-abs"
```
From the borg-ui host (mount options mirror `build_sshfs_command`):
```sh
# current behavior
sshfs root@REMOTE:$d /mnt -o follow_symlinks -o workaround=rename
# faithful behavior for symlinks
sshfs root@REMOTE:$d /mnt -o no_contain_symlinks -o workaround=rename
borg create ::a /mnt && (cd /empty && borg extract ::a) && ls -l /empty && stat -c '%h %i %n' /empty/*
```

## Proposed remediation

1. **Backup-source mount → `-o no_contain_symlinks` and *not* `follow_symlinks`.** This restores restore = backup for symlinks. I'd argue this should **not** be user-configurable: "the restore reproduces the source" is an invariant of a backup tool, not a toggle. (If a deliberate "dereference targets" mode is ever wanted, it can be added later as an explicit opt-in.)
2. **Parameterize the shared `build_sshfs_command()`** so only the backup call gets the preserve options; browse/cloud-mirror and restore keep their current behavior (no regression). This is a small, self-contained change.
3. **Document the hard-link limitation.** It cannot be fixed over sshfs; the faithful path for hard links is running borg on the source host (e.g. an agent / `ssh://` + remote borg). A note in the SSH-source docs/changelog is the honest fix.
4. **(Separate cleanup)** Remove the dead, unreferenced legacy builder `_mount_ssh_path_OLD_BROKEN()` / `_unmount_ssh_path_OLD()` (`backup_service.py:897` / `:1133`, marked "DO NOT USE"), which contains a second hardcoded `follow_symlinks` and could otherwise be missed when the live one is fixed.

## Open design question — browsing

Interactive browsing of a remote source arguably wants `follow_symlinks` for UX (click a link, see the target). But absolute symlinks that escape the mount root cannot be transparently followed under either option set (contain blocks them; `no_contain` yields a client-side-absolute target that resolves on the wrong host). A dedicated browse mounter with an explicit UI caveat ("absolute symlinks outside the mounted path are not traversable") seems cleaner than sharing the backup mounter. Would appreciate maintainer input on whether to split the browse path out — happy to do it in a follow-up PR.

---

I'm happy to open PRs for (1)+(2)+(3) and the (4) cleanup separately. Wanted to align on direction — and surface any intentional reason `follow_symlinks` was chosen — before implementing.

ioanalytica:
Fixes #751 end to end (backup **and** restore).

> **⚠️ Note for release notes.** This changes fidelity, but it cannot repair the past: SSHFS-source archives created **before** this fix already have their symlinks dereferenced and their absolute/broken links dropped — only new backups are faithful. Worth calling out so users know to re-check important SSHFS-source archives (the usage guide explains how: `borg list` and look for symlinks stored as regular files, or compare a test restore against the source). The auto-generated changelog uses commit subjects only, so this note won't surface automatically.

Both a backup source and a restore destination reached over SSH are mounted with `sshfs` via the same `build_sshfs_command()`, which hardcoded `-o follow_symlinks`. That single option breaks fidelity in **both** directions. Both were reproduced empirically on a real sshfs mount to a Linux remote (kernel 6.8, borg 1.4.4), measuring the ground truth directly on the remote filesystem.

## Problem 1 — backup (read side): symlinks are mangled before Borg sees them

Borg does `lstat` and never follows symlinks itself, so whatever the sshfs mount presents is what gets archived. With `follow_symlinks`, sshfs dereferences on read; and simply turning it *off* is not enough either — sshfs's `contain_symlinks` sandbox (on by default in sshfs 3.x) makes `readlink` on an out-of-root target fail with `EPERM`, so Borg logs `E` and drops absolute links.

Restore compared to the live source, per mount option:

| source entry | `follow_symlinks` (old) | `no_contain_symlinks` (fix) |
|---|---|---|
| relative symlink | **dereferenced** to a file | stored as the symlink |
| symlink to a dir | target tree **duplicated** | stored as the symlink |
| absolute symlink (`→/etc/hostname`) | **dereferenced** | stored with its target |
| broken symlink | **dropped** | stored with its target |

`borg create`/`borg check` cannot detect this — from Borg's view it faithfully stored what the filesystem presented.

## Problem 2 — restore (write side): `borg extract` errors on every symlink

`borg extract` into an sshfs-mounted destination creates each link and then sets its attributes. Under `follow_symlinks` the mount dereferences the just-created link, so the attr step follows to the (often missing) target and fails. Ground truth measured on the remote after extract:

| archived entry | `follow_symlinks` (old restore default) | `no_contain_symlinks` (fix) |
|---|---|---|
| symlinks (rel / dir / abs / broken) | land correctly, but `borg extract` reports `attrs: [Errno 5] I/O error` (broken: `Errno 2`) on **every** symlink | land correctly, **clean** |
| hard links | recreated (same inode) | recreated (same inode) |
| extract exit | non-zero — restore reported as failed/partial | clean |

So the data lands either way, but `follow_symlinks` makes every SSH restore that contains a symlink report failure — noisy and misleading, and it can mask real errors. `no_contain_symlinks` does **not** block link *creation* (contain only guards read traversal), so absolute and broken links extract cleanly.

## Change

Mount both **backup sources** and **restore destinations** with faithful symlink handling; leave browse and cloud-mirror untouched.

- New `mount_service._sshfs_symlink_options(preserve_symlinks)`:
  - `True` → `-o no_contain_symlinks` (and **not** `follow_symlinks`) — the fidelity path.
  - `False` (default) → `-o follow_symlinks`, exactly as before.
- `_execute_sshfs_mount()` and `mount_ssh_paths_shared()` take `preserve_symlinks` (default `False`).
- `backup_service` (source mount) and `restore_service` (destination mount) pass `preserve_symlinks=True`. Browse/cloud-mirror (`mount_ssh_directory`) keep `follow_symlinks` — a UX choice for interactive browsing, not a fidelity path — so there is no regression there.

### Hard links

Note the asymmetry, both proven above: a **backup** cannot capture hard links over SSHFS — SFTP v3 transmits no `st_ino`/`st_nlink`, so sshfs reports `nlink=1` with synthetic per-path inodes and Borg never enters hard-link detection. A **restore**, however, recreates them faithfully (same inode on the remote) via SFTP's `hardlink@openssh.com` extension. `docs/usage-guide.md` documents the backup-side limitation and the faithful alternative (back the source up from the machine itself with a managed agent).

## Tests

- `_sshfs_symlink_options` returns `no_contain_symlinks` (preserve) / `follow_symlinks` (default), mutually exclusive.
- `_execute_sshfs_mount(preserve_symlinks=True)` builds an argv with `no_contain_symlinks` and **not** `follow_symlinks`; the default still emits `follow_symlinks`.
- `backup_service` mounts SSH sources, and `restore_service` mounts SSH destinations, with `preserve_symlinks=True`.

No frontend changes.

## Summary by CodeRabbit

* **New Features**
  * SSH-based backups and restores now preserve symbolic links, including absolute and broken links, for more faithful file structures.
  * Existing browsing and cloud-mirroring behavior remains unchanged.

* **Documentation**
  * Added guidance on remote source fidelity, hard-link limitations, disk usage considerations, and the impact on previously created archives.

### DIFF
--- a/app/services/backup_service.py
+++ b/app/services/backup_service.py
@@ -1021,6 +1021,10 @@ class BackupService:
                         "connection_id": connection_id,
                         "remote_paths": remote_paths,
                         "job_id": job_id,
+                        # Backup sources must reproduce the source on restore: mount
+                        # with faithful symlink handling (no dereference, keep
+                        # absolute/broken links). See mount_service._sshfs_symlink_options.
+                        "preserve_symlinks": True,
                     }
                     if shared_ssh_temp_root is not None:
                         mount_kwargs["temp_root"] = shared_ssh_temp_root
--- a/app/services/mount_service.py
+++ b/app/services/mount_service.py
@@ -64,6 +64,30 @@ def _decrypt_with_module_secret(encrypted_value: str) -> str:
     return cipher.decrypt(encrypted_value.encode()).decode()
 
 
+def _sshfs_symlink_options(preserve_symlinks: bool) -> list[str]:
+    """SSHFS options controlling how symlinks in the remote tree are presented.
+
+    ``preserve_symlinks=True`` is the fidelity path, used for both directions:
+
+    - Backup source (read): the archive must reproduce the source, so do NOT
+      dereference symlinks and keep absolute/broken ones. Without this,
+      ``follow_symlinks`` dereferences them and sshfs's ``contain_symlinks``
+      sandbox (on by default in sshfs 3.x) drops absolute links (``readlink``
+      EPERM). Safe: borg in nofollow mode only reads the target string.
+    - Restore destination (write): ``borg extract`` creates the links, then sets
+      their attributes. Under ``follow_symlinks`` the mount dereferences each new
+      link so the attr step follows to the (often missing) target and fails with
+      an I/O error on every symlink. ``no_contain_symlinks`` avoids that; it does
+      NOT block link *creation*, so absolute/broken links extract cleanly.
+
+    Only browse and cloud-mirror keep the historical ``follow_symlinks`` — a UX
+    choice for interactive browsing, not a fidelity path — via ``preserve_symlinks=False``.
+    """
+    if preserve_symlinks:
+        return ["-o", "no_contain_symlinks"]
+    return ["-o", "follow_symlinks"]
+
+
 def _sshfs_missing_remote_path(error_message: str) -> bool:
     normalized = (error_message or "").lower()
     return any(marker in normalized for marker in SSHFS_NO_SUCH_FILE_MARKERS)
@@ -535,6 +559,7 @@ class MountService:
         remote_paths: List[str],
         job_id: Optional[int] = None,
         temp_root: Optional[str] = None,
+        preserve_symlinks: bool = False,
     ) -> Tuple[str, List[Tuple[str, str]]]:
         """
         Mount multiple remote SSH directories from the same connection under a single shared temp root.
@@ -545,6 +570,8 @@ class MountService:
             remote_paths: List of remote paths to mount (all from the same connection)
             job_id: Optional backup job ID for tracking
             temp_root: Optional existing temporary root to reuse across source connections
+            preserve_symlinks: Mount for faithful symlink fidelity (backup sources
+                and restore destinations); browse/cloud-mirror keep follow_symlinks
 
         Returns:
             Tuple of (temp_root, mount_info_list)
@@ -782,6 +809,7 @@ class MountService:
                         remote_path=mount_remote_path,
                         mount_point=mount_dir,
                         temp_key_file=temp_key_file,
+                        preserve_symlinks=preserve_symlinks,
                     )
 
                     # Verify mount with READ-ONLY check (NEVER write to user data!)
@@ -1587,8 +1615,14 @@ class MountService:
         remote_path: str,
         mount_point: str,
         temp_key_file: str,
+        preserve_symlinks: bool = False,
     ):
-        """Execute SSHFS mount command with SSH key authentication"""
+        """Execute SSHFS mount command with SSH key authentication.
+
+        ``preserve_symlinks`` selects faithful symlink handling for backup sources
+        and restore destinations (see ``_sshfs_symlink_options``); browse and
+        cloud-mirror keep the default ``follow_symlinks`` behavior.
+        """
         sftp_server_path = None
         use_sudo = getattr(connection, "use_sudo", False)
 
@@ -1673,8 +1707,7 @@ class MountService:
                 "ServerAliveCountMax=3",
                 "-o",
                 "reconnect",
-                "-o",
-                "follow_symlinks",
+                *_sshfs_symlink_options(preserve_symlinks),
                 "-o",
                 "allow_other",
                 "-o",
--- a/app/services/restore_service.py
+++ b/app/services/restore_service.py
@@ -1259,6 +1259,12 @@ class RestoreService:
                 connection_id=destination_connection_id,
                 remote_paths=[destination],
                 job_id=job_id,
+                # Extract must write faithful symlinks. With follow_symlinks the
+                # mount dereferences each just-created link, so borg's attr step
+                # follows to the (often missing) target and fails with an I/O
+                # error on every symlink. no_contain_symlinks avoids that and
+                # keeps absolute/broken links intact. See _sshfs_symlink_options.
+                preserve_symlinks=True,
             )
 
             if not mount_info_list:
--- a/docs/usage-guide.md
+++ b/docs/usage-guide.md
@@ -175,6 +175,32 @@ Then use `/documents`, `/photos`, and `/backups` in the UI.
 
 For remote sources, choose a Remote Machine in the Backup Plan and select paths from that machine.
 
+### Remote (SSH) source fidelity
+
+When a source is reached over SSH, Borg UI mounts it with `sshfs` and runs Borg
+over the mount. Two things are worth knowing about how the restore compares to
+the live source:
+
+- **Symbolic links are preserved as-is** — including absolute and broken links —
+  so a restore reproduces the source's link structure. Borg UI mounts backup
+  sources without dereferencing symlinks.
+- **Hard links are not preserved when backing up over SSHFS.** The SFTP protocol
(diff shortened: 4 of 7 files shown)


karanhudia (approved):
Reviewed the SSHFS symlink-fidelity behavior, documentation, and restoration coverage; approved for merge.

PR #765, opened 2026-07-19
fix(agent): make the session thread the only writer on the WebSocket

ioanalytica:
## Problem

A managed agent connected over **`wss://`** loses its session whenever a command produces a larger response. The UI reports `502 backend.errors.agents.filesystemBrowseFailed` / *"Agent session disconnected"*, the agent logs `Connection to remote host was lost.` and reconnects ~2 s later. Agents on plain `ws://` are unaffected.

Measured against one deployment (identical call, only the path differs):

| Transport | Path | Payload | Result |
|---|---|---|---|
| `ws://` (in-cluster) | `/usr/bin` | 23 KB | 200, stable |
| `wss://` (ingress) | `/var` | 1.5 KB | 200, 3/3 |
| `wss://` (ingress) | `/etc` | 27 KB | 502 in ~2 of 3 runs |

This is not an ingress or network fault. Path MTU is clean, 10 × 64 KB HTTPS POSTs over the exact same path succeed 10/10 in ~50 ms, idle sessions survive 33 min, and the WS handshake negotiates no compression extension. `tcpdump` on the agent host shows the **agent** sending FIN first on healthy TCP — no RST, no timeout, no close from the peer. It dies while *sending* the large response.

## Root cause

`run_session()` deliberately stays in `socket.recv()` so websocket-client auto-pongs the server's keepalive pings while a job runs (#707). `_handle_command` therefore runs in worker threads — and those threads call `socket.send()` directly for `command_ack`, `job_started`, `log`, `progress` and ephemeral `command_result`. The existing `_send_lock` serializes senders against each other, but **not against `recv()`**.

websocket-client uses two separate locks for exactly these two paths (`_core.py`: `self.lock` for send, `self.readlock` for recv), so a send and a recv genuinely overlap on the same socket. On a raw TCP socket that is harmless. On an `ssl.SSLSocket` it is not — the OpenSSL object is not thread-safe, and concurrent use corrupts its state, producing an SSLError / unexpected EOF that surfaces as `WebSocketConnectionClosedException`.

This explains every observation: idle sessions are stable (the heartbeat is sent *from* the recv loop, never concurrently), only responses from worker threads trigger it, it is size-dependent but not deterministic (the race window grows with the number of `SSL_write` calls), and `ws://` agents never see it.

## Relation to #735

#735 moved terminal job outcomes to the REST job API and described the cause as a large frame holding websocket-client's send lock long enough to starve the keepalive pong. That change is worth keeping on its own merits — `/complete`, `/fail` and `/cancel` are idempotent, so a result survives a session drop instead of orphaning the job for 15 minutes until the reaper.

But the diagnosis was incomplete, which is why the symptom came back for TLS agents. The problem is not how *long* a worker holds the send lock, it is that a worker touches the socket **at all** while the loop is in `recv()`. #735 removed one of five worker send paths (terminal results); `command_ack`, `job_started`, `log`, `progress` and ephemeral `command_result` still raced. Frame size only scales the race window — a log-streaming backup is exposed to the same failure, just less visibly than a 117 KB browse response.

So this PR does not revert #735, it generalizes it: instead of moving individual payloads off the socket one at a time, no worker thread writes to the socket at all. I have updated the `_deliver_terminal` docstring accordingly, keeping the reasons for the REST path that still hold (size, idempotency, surviving a drop) and dropping the send-lock/pong explanation.

## Fix

Single-writer model: the session thread is the only one that ever touches the socket.

- Workers hand outgoing frames to a `queue.Queue`; the session loop drains it between `recv()` calls.
- `SessionCommandClient` no longer holds a socket reference at all, so the invariant is enforced by construction rather than by convention.
- The recv timeout is shortened to 1 s so queued frames are not held back, with the heartbeat/ping kept on its original 30 s cadence via a monotonic clock.
- Flushing temporarily restores the full session timeout, because `settimeout` bounds **writes** as well as reads and a result frame can be hundreds of KB.
- The outbox is bounded and drops with a warning instead of growing without bound if the socket stops draining.

A shared lock over both send and recv is not an alternative: the blocking `recv()` would lock every sender out.

The wire protocol is unchanged, so the server side needs no change and no coordinated rollout — only a docstring recording that `AgentConnectionManager` state is process-local, which makes a scaled-out deployment reproduce the same 502s with no network fault involved.

## Testing

- New regression test asserts the invariant directly: the job handler really runs off-thread, yet every `send()` came from the session thread. It also covers frame ordering and the timeout switch.
- Full backend unit suite passes (2508 passed), `ruff check` / `ruff format --check` clean.
- Verified in a real deployment: after rolling the updated agent, filesystem browse over `wss://` succeeds 100 %, including a full backup run.

## Summary by CodeRabbit

* **Improvements**
  * Improved real-time command delivery using a single-writer WebSocket sending model with ordered, bounded outbox buffering.
  * Enhanced keepalive and reconnect/shutdown behavior, including safer handling during connection drops.
  * Clarified how persisted job outcomes are delivered versus temporary command responses.
* **New Features**
  * Added queued command enqueueing to safely handle concurrent command updates.
* **Bug Fixes**
  * Prevented inconsistent delivery during simultaneous command completion scenarios.
* **Tests**
  * Updated and expanded runtime tests to verify outbox draining, ordering, and send/timeout semantics.
* **Release**
  * Updated version to **0.1.2**.


ioanalytica:
Thanks — both points addressed in `88af555c`, in each case with a slightly wider scope than suggested. Summary of where I deviated and why:

**1. Hello frame under the short poll timeout (Major).** Correct, and the same exposure exists one function over: `_send_keepalive` writes the heartbeat under the very same 1 s budget. Moving the `_set_socket_timeout` call after `_send_hello` fixes the reported instance but leaves the class open. Instead, all three write paths (outbox flush, hello, keepalive incl. the protocol ping) now run inside a `_writing()` context manager that raises the timeout for the write and drops back to the poll interval afterwards. `_flush_outbox` keeps its batch semantics — one toggle per flush, not per frame.

**2. Silent frame loss on the `outbox` fallback (Nitpick).** Taken, but a warning about a frame that will never be delivered is still a silent drop from the caller's perspective. `outbox` is now a required parameter with no default, so omitting it is a `TypeError` at construction. That matches what the rest of the PR does — `SessionCommandClient` also lost its socket reference entirely rather than being documented as "don't use this".

**3. The test gap you noted was the most useful part.** `FakeWebSocket.send()` did not enforce timeouts, so this would have sailed through the suite. The fake now records the timeout in force for every write, and two tests assert `set(send_timeouts) == {30}`: the single-writer invariant test and the idle-heartbeat test, i.e. exactly the two paths involved.

Also included, from a review on our fork of the same branch: the `AgentConnectionManager` docstring said the constraint holds "with a single application replica", which is wrong — the state is per *process*, and `entrypoint.sh` pins `gunicorn --workers 1`. Raising the worker count breaks it just as adding replicas would, so the docstring now names that.

Full backend unit suite passes (2508), `ruff check` / `ruff format --check` clean.

karanhudia (approved):
Reviewed the single-writer WebSocket design, addressed timeout feedback, and regression coverage; approved for merge.

PR #772, opened 2026-07-23
feat: unified retention for job logs, job history, and pruned archives

ioanalytica:
## Problem

Log files on disk have retention (an age window, a total-size cap) and a save policy — the database has none of it:

- Every agent log line becomes an `agent_job_logs` row. On real installs this table is **~90% of all database rows**, and nothing ever prunes it.
- Every finished job keeps its row and often an inline `logs` copy forever, across all job tables.
- `log_save_policy` claims to control which jobs keep their logs, but is only applied on the read side — the database stores everything regardless.
- On SQLite the database file grows without bound.

## What this PR does

**Commit 1 — unified retention.** The database now follows the same rules as the files:

- `log_save_policy` applies to DB-stored log content exactly as to files: outcomes the policy does not keep (clean successes under `failed_and_warnings`, also warnings under `failed_only`) lose their log content at the next pass, regardless of age.
- `log_retention_days` bounds DB log content the way it bounds files: agent job log rows are deleted, inline `logs` columns cleared once the window passes.
- `cleanup_retention_days` bounds the job records themselves — every kind of job row including plan runs and script executions, **regardless of status**. Age is taken from the freshest timestamp a table has, so live work never looks old while zombies (e.g. queued for an agent that never returned) finally fall. Retention is a property, not an option: there is no off switch, only the window moves.
- Applied by a daily background loop and by the previously-placeholder `POST /api/settings/system/cleanup` endpoint, which also VACUUMs SQLite so the file actually shrinks. Deletes are chunked; dialect-agnostic (SQLite and Postgres).
- UI: the Log Management tab gains the job-history window next to the log window on the same 7–90 day scale, the whole tab (previously hardcoded English prose) now goes through i18n with translations for all four locales, and a Storybook story covers the tab.

**Commit 2 — pruned archives take their job records with them.** A backup job whose archive was pruned is not history anymore but a reference into the void. When borg removes an archive, the records that produced it now go too: the log file on disk, the linked agent job (log rows and script executions fall via the DB cascades), and the `backup_jobs` row.

- The server prune path parses the `Pruning archive:` lines of the `--list` output it already produces; agent prunes now pass `--list` as well (agent 0.1.3); archive deletes know the name directly.
- Agent log ingestion races the command result — at completion, most log lines have not arrived yet (observed live: 1 of 30), which also left maintenance job logs truncated in the UI. A sweep in the retention pass re-parses the complete agent logs later, cascades idempotently, and repairs the truncated stored logs.
- Borg 2 repositories are skipped for now: an archive *series* shares one name across archives, so a name match would hit jobs whose archives still exist. Lifting that needs the archive id captured per job.

## Testing

- 12 unit tests for the retention service: both windows, the policy phase, status-independent row purge with live-work protection, the cascade (including a verbatim borg 1.4 prune line and dry-run negatives), the race sweep with log repair, idempotency.
- Frontend: typecheck, eslint, prettier, locale parity, Storybook build; the story pins the out-of-range clamp.
- Validated end to end on two real installs (one SQLite, one Postgres): retention windows and policy phase with seeded fixtures, and the prune cascade against a live prune round — 8 job records cascaded, 30 truncated prune logs repaired.

## Summary by CodeRabbit

* **New Features**
  * Added configurable job-history retention alongside log retention.
  * Added background job-history retention scheduling and improved manual cleanup execution.
  * Added automatic cleanup behavior in Log Management (including history-retention controls).
  * Added translations for expanded log-management UI.
* **Bug Fixes**
  * Job-history records are now purged after completed prune and delete operations, including pruned-archive handling.
  * Cleanup now performs real retention work and reports results (including optional SQLite vacuum).
  * Updated agent version to **0.1.3** and borg prune commands to include `--list`.


ioanalytica:
All four findings addressed in place:

- `db.commit()` releases the request session's read transaction before the SQLite `VACUUM` runs on its own connection.
- Log files of pruned archives are now unlinked only after the DB delete has committed (paths captured beforehand), so a failed commit can no longer leave rows pointing at vanished files.
- `cleanup_retention_days ?? 90` so a stored `0` clamps to the 7-day minimum instead of silently becoming 90.
- Spanish retention description reworded as suggested.

@coderabbitai review

karanhudia (approved):

PR #796, opened 2026-08-07
fix(db): two migration edge cases

ioanalytica:
Two edge-case fixes to the one-time upgrade from #768.

- **Don't re-apply legacy migrations on an already-migrated database** — an
  Alembic SQLite moved into a fresh Postgres re-ran them; skip when the source
  already carries `alembic_version`.
- **Leave `borg.db` in place when migrating to Postgres** — renaming it to `_bak`
  was pointless (Postgres is the live database) and collided with a `_bak` a prior
  in-place upgrade left behind.

One commit each, with tests.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved database upgrades when moving from SQLite to PostgreSQL.
  - PostgreSQL upgrades now preserve the original SQLite database and avoid creating unnecessary rollback files.
  - SQLite upgrades now create a rollback backup before replacing the active database.
  - Existing migration metadata is preserved, preventing already-upgraded databases from rerunning legacy migrations.

- **Tests**
  - Added coverage for PostgreSQL finalization, SQLite rollback handling, and complete database transfers.

### DIFF
--- a/app/database/db_upgrade.py
+++ b/app/database/db_upgrade.py
@@ -13,7 +13,8 @@ done before the copy runs -- copying by column name cannot reproduce them. Any
 user can upgrade straight from an old release to the latest; no manual stop on an
 intermediate version is required.
 
-Nothing is destroyed. The source is renamed, never deleted, and every deviation
+Nothing is destroyed. The source is kept as the rollback -- moved aside for a
+SQLite upgrade, left exactly where it is for a Postgres one -- and every deviation
 between source and target is reported rather than silently absorbed.
 """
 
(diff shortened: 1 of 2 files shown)


ioanalytica:
The failing **Backend Security (pip-audit)** check here is not from this change — this PR touches only `app/database/db_upgrade.py` and its tests, no dependencies. It is a repo-wide advisory (`cryptography` 49.0.0 → **PYSEC-2026-3552**, fixed in 50.0.0) that fails on `main` and every open PR alike.

#797 clears it (and the parallel npm-audit advisory, `dompurify` GHSA-55q2-fjhq-7xh7). Once #797 is merged, this PR's checks go green on a re-run/rebase.

ioanalytica:
@karanhudia: May I bring in the updates from main via a rebase rather than a merge?

karanhudia (approved):
Reviewed the SQLite-to-Postgres upgrade edge-case handling and its targeted regression tests. Full CI, smoke tests, and security checks pass.

PR #851, opened 2026-08-24
fix(dashboard): surface prune, scheduled check, and restore runs in the activity timeline

ioanalytica:
## Problem

The dashboard's 14-day activity timeline renders five lanes (backup, check, compact, restore, prune), but three of them can never show anything:

1. **Prune lane always empty** — `GET /api/dashboard/overview` builds the activity feed from `BackupJob`, `CheckJob`, `CompactJob`, and `RestoreCheckJob` only; `PruneJob` is never queried, so completed prunes are simply not fetched.
2. **Scheduled checks and compacts invisible** — the Borg 2 check and compact services never set `started_at`. The manual endpoints pre-create their job as `running` with `started_at` set, but scheduler-dispatched jobs arrive as `pending`, so they complete with `started_at = NULL` and fall out of every `started_at`-based view — including the activity feed's 14-day window (`started_at >= now-14d`).
3. **Restore lane dead** — the API emits `type: "restore_check"`, but the timeline matches lane names exactly (`LANES.indexOf(a.type)`), so restore verification events are silently dropped.

Observed on a real install: nightly scheduled checks (61 failures in 14 days among them) and ~160 prunes/day were invisible while the dashboard reported "All systems nominal".

## Fix

- `app/api/dashboard.py`: query recent `PruneJob` rows and emit `type: "prune"` feed entries, following the existing per-type pattern.
- `app/services/v2/check_service.py`, `app/services/v2/compact_service.py`: mark the job `running` and record `started_at` at execution start, mirroring the v1 check and v2 prune services. The start transition is a conditional claim (`status IN ('pending', 'running')`, proceed only when a row matched), so a concurrent cancellation is never overwritten back to `running`.
- `frontend/.../ActivityTimeline.tsx`: map `restore_check` events onto the `restore` lane (position and color).

Complements #850: that fix routes scheduled checks to the executor that should run them; this one makes their runs (and prunes/restore checks) visible on the dashboard.

## Tests

- Overview test seeds a `PruneJob` and asserts the full feed type sequence.
- V2 service tests pin the scheduler path (a `pending` job ends `completed` **with** `started_at` set) and the cancellation race (a job cancelled between the status guard and the start claim stays `cancelled`, borg is never spawned) for check and compact.
- Frontend test: a `restore_check` event renders on the restore lane in the restore color.

### DIFF
--- a/app/api/dashboard.py
+++ b/app/api/dashboard.py
@@ -19,6 +19,7 @@ from app.database.models import (
     ScheduledJobRepository,
     CheckJob,
     CompactJob,
+    PruneJob,
     RestoreCheckJob,
     SSHConnection,
     SystemSettings,
@@ -1044,6 +1045,9 @@ async def get_dashboard_overview(
             .filter(CompactJob.started_at >= fourteen_days_ago)
             .all()
         )
+        recent_prunes = (
+            db.query(PruneJob).filter(PruneJob.started_at >= fourteen_days_ago).all()
+        )
         recent_restore_checks = (
             db.query(RestoreCheckJob)
             .filter(RestoreCheckJob.started_at >= fourteen_days_ago)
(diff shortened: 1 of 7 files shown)


ioanalytica:
@coderabbitai review

PR #855, opened 2026-08-25
feat(agents): offer Borg verify/update in the reinstall dialog

ioanalytica:
## What

The **Reinstall agent** dialog only emitted `--reinstall`, which updates the agent package but never touches Borg — although `install.sh` already supports `--reinstall --borg-version 1|2|both` to also verify or update the Borg binaries to the versions the server provides. Updating an installer-managed Borg therefore required knowing the flags by heart.

The dialog now shows the same Borg selection cards as the add-agent wizard, **preselected from what the agent last reported**, and the generated command carries `--borg-version` accordingly.

## How

- **`BorgInstallModeRadioGroup`** (new): the selection cards extracted from `AddAgentDialog`. The two contexts describe the same choices differently (install vs verify/update, and which option is the default — in reinstall mode that's *Skip*), so the labels come from the caller's i18n prefix.
- **`deriveBorgInstallMode(borg_versions)`**: the install flags themselves are not persisted anywhere, but the agent classifies each detected binary's `install_source`, and `borg-ui-installer` means the binary lives where install.sh puts the Borgs it manages. Majors reported with that source preselect `1`/`2`/`both`; system packages, custom paths, or agents without a report map to the reinstall default of leaving Borg untouched.
- **`buildAgentReinstallCommand(url, mode)`**: appends `--borg-version`; *Skip* stays flag-free (reinstall mode skips Borg by default). The selection can be overridden either way.
- **Info alert**: now also notes that the reinstall preserves the service user recorded in the systemd unit, so the command can be run from any sudo-capable account — the operator does not need to remember the original install user.

UI-only; no backend changes.

## Tests

- Builder matrix (`skip`/`borg1`/`borg2`/`both`) and derivation matrix (installer-managed majors, system-package, custom-path, empty, undefined).
- Dialog interaction test: preselects *both* from an installer-managed report, copy yields `--borg-version both`, switching to *Skip* drops the flag.
- Existing reinstall tests unchanged and green; full suite 2313/2313.
- Storybook: new `AgentReinstallDialogInstallerManagedBorg` story shows the preselected state.
- Locales: en/de/es/it in parity.

Reviewed by CodeRabbit on our fork (ioanalytica#34, all threads resolved) before submission.

### DIFF
--- a/frontend/src/locales/de.json
+++ b/frontend/src/locales/de.json
@@ -341,7 +341,28 @@
         "title": "Agent neu installieren",
         "copyCommand": "Neuinstallationsbefehl kopieren",
         "description": "Führen Sie dies auf dem registrierten Rechner aus, um das Borg-UI-Agent-Paket zu aktualisieren und den systemd-Dienst neu zu starten. Kein Registrierungstoken oder Registrierungsschritt erforderlich.",
-        "configPreserved": "Das Skript behält die vorhandene Agent-Konfiguration bei unter"
+        "configPreserved": "Das Skript behält die vorhandene Agent-Konfiguration unter",
+        "serviceUserPreserved": "Der Dienst läuft außerdem weiter unter dem in seiner systemd-Unit hinterlegten Benutzer. Führen Sie den Befehl von einem beliebigen sudo-fähigen Konto aus.",
+        "borgInstallation": "Borg-Installation",
+        "borgSelectionHint": "Vorausgewählt anhand der letzten Meldung des Agents. Eine Borg-Auswahl aktualisiert diese Binaries zusätzlich auf die Versionen dieses Servers.",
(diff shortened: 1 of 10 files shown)


karanhudia:
The implementation is clean, but I do not want a normal agent reinstall to preselect a Borg upgrade. That can unexpectedly move users between Borg versions, especially while Borg 2 is still changing repository formats. Please default to Skip and make a Borg change an explicit user choice, with clear upgrade wording.

ioanalytica:
Done in 8baeb6dd — the preselection is removed entirely. The dialog always starts at Skip (and resets to Skip when it targets another agent), a Borg change is an explicit selection, and the hint now states plainly that selecting Borg versions upgrades those binaries to the versions this server provides. Branch rebased onto current main.

PR #870, opened 2026-09-02
fix(agents): send notifications for agent-executed jobs

ioanalytica:
## Problem

The notification settings expose per-event toggles (backup start/success/warning/failure, check success/failure), but none of them ever fire for repositories executed by a managed agent. The server-side execution paths send their notifications from inside `backup_service` / `check_service`; agent-executed jobs never pass through those services — they reach their terminal state in the agent transport handlers (`app/api/agents.py`, WebSocket + REST) and in the agent job reaper, none of which called the notification service. A failing agent backup or check was therefore silent, even with failure notifications enabled.

## Changes

- New `app/services/agent_job_notifications.py`: renders the linked `BackupJob` / `CheckJob` rows into the same `notification_service` calls the server-side paths make. The job name for notification titles resolves via `backup_plan_id` (plan name) or `scheduled_job_id` (schedule name); naive DB datetimes are coerced to UTC before duration math. Notification failures are logged and never propagate into the agent protocol.
- Both transports dispatch after the terminal state is committed (WebSocket `command_result` / `command_error`, REST `/complete`, `/fail`). Terminal transitions are claimed atomically: a shared guarded conditional UPDATE (`_claim_terminal_transition`, WHERE status not final — the same pattern the job reaper already uses) makes exactly one writer win, so a redelivered report (e.g. WebSocket plus a REST retry) or a concurrent reaper pass can neither double-notify nor overwrite an already-final state. This includes completion reports the server classifies as failures (borg error exit codes, malformed return codes), and the first backup-start notification is claimed the same way (guarded UPDATE on `started_at IS NULL`).
- The backup-start notification fires on the job's first start report (WebSocket + REST); a requeued job keeps its original `started_at` and does not notify again.
- The whole notification dispatch — including every ORM attribute read, which can itself hit the database on a committed (expired) object — runs inside an exception boundary: a failure is logged and never propagates into the agent protocol. The exception handlers log primary keys via an identity-map helper (`orm_identity_id`) so the logging itself cannot re-raise on an expired object.
- The REST `/start`, `/fail` and `/cancel` endpoints now reuse the shared transition helpers instead of duplicating the logic inline, so a late cancel report cannot overwrite a job the reaper or the other transport already finalized.
- Agent-executed repository checks (`repository.check` with a linked `CheckJob`) send `send_check_completion` on success and failure, with `check_type` derived from `scheduled_check`.
- The agent job reaper reports the backup jobs it fails (agent session lost) through the same path: `reap_stale_agent_jobs` collects the flipped backup-job ids via an optional out-parameter (signature otherwise unchanged) and the background loop sends the failure notifications with its own session.

Deliberate parity choices with the server-side paths: cancelled jobs do not notify; a check that completed with warnings is reported as completed (the repository was verified); prune/compact remain notification-less (no toggles exist for them).

## Tests

17 new tests in `tests/unit/test_api_agents.py`: success/warning/failure notifications over both transports, exactly-once delivery on repeated completion reports, start-notification dedup, check success/failure, plan-name resolution, cancelled-status skip, reaper collection (positive case + terminal-status guard), stale-transport races for completion, start and cancel (a second session models the other transport losing the guarded claim), and the exception boundary (a detached/expired row and a raising notification service must not propagate). The full file passes, the adjacent agent/notification/backup/repository suites stay green, and `ruff check` / `ruff format` are clean.

### DIFF
--- a/app/api/agents.py
+++ b/app/api/agents.py
@@ -49,6 +49,12 @@ from app.services.agent_job_dispatcher import (
     agent_job_kind as live_agent_job_kind,
     dispatch_agent_job_best_effort,
 )
+from app.services.agent_job_notifications import (
+    notify_backup_job_finished,
+    notify_backup_job_started,
+    notify_check_job_finished,
+    orm_identity_id,
+)
 from app.utils.datetime_utils import serialize_datetime
 
 logger = structlog.get_logger()
@@ -527,9 +533,12 @@ def _mark_agent_job_claimed(job: AgentJob, *, now: Optional[datetime] = None) ->
 
 def _mark_agent_job_started(
     job: AgentJob, db: Session, *, started_at: Optional[datetime] = None
-) -> None:
+) -> Optional[BackupJob]:
+    """Returns the linked backup job when this start report is its first one,
+    so the caller can send the backup-start notification exactly once (a
+    requeued job keeps its original started_at and does not notify again)."""
     if job.status in FINAL_AGENT_JOB_STATUSES:
-        return
+        return None
     now = _now_utc()
     if job.claimed_at is None:
         job.claimed_at = now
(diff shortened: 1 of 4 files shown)


karanhudia:
Good catch 💯

PR #873, opened 2026-09-02
fix(activity): apply status filter before the row limit

ioanalytica:
## Problem

The recent-activity endpoint (`/api/activity/recent`) fetches the newest `limit` rows per lane and only then drops non-matching statuses in Python. Filtering by status therefore searches *within* the newest rows instead of returning the newest *matching* rows. On a busy instance (hourly plans across a dozen repositories produce hundreds of entries per day) a failed backup slides out of the newest-200 window within hours — after that, `status=failed` returns nothing, although the job row and its logs still exist. The Activity page then has no way to reach the failure at all: the filter is forwarded to the API, but the API applies it after the cut.

## Changes

The status filter becomes part of each lane's SQL query, before ordering and limit, so `status=failed` returns the newest failed jobs regardless of how many successes came after them. Applied uniformly to all eight status-carrying lanes (backup, restore, check, restore-check, compact, prune, script executions, package installs, rclone). The two availability-skip lanes, whose entries carry the synthetic status `skipped`, skip their queries entirely for any other status filter instead of discarding every row afterwards. Behavior without a status filter is unchanged, as is the combined sort-and-truncate of the merged result.

## Tests

Two new tests in `tests/unit/test_api_activity.py`: a failed backup buried under more than `limit` newer successes is returned by `status=failed` (fails on the previous code), and `status=skipped` still returns the availability-skip lanes after their new lane-level guard. Full activity suite green; ruff check and format clean.

### DIFF
--- a/app/api/activity.py
+++ b/app/api/activity.py
@@ -329,12 +329,15 @@ async def list_recent_activity(
 
     # Fetch backup jobs
     if not job_type or job_type == "backup":
+        # Filter in SQL, before the limit: "status=failed" must return the
+        # newest failed jobs, not the failed jobs among the newest rows.
+        backup_query = db.query(BackupJob)
+        if status:
+            backup_query = backup_query.filter(BackupJob.status == status)
         backup_jobs = (
-            db.query(BackupJob).order_by(BackupJob.started_at.desc()).limit(limit).all()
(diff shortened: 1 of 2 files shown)


karanhudia:
Good catch 💯

PR #875, opened 2026-09-02
fix: report repository size for Borg 2 agent repositories

### LINKED ISSUE #874: Borg 2 agent repositories always show "Total size: N/A"
All my agent-backed repositories show `Total size: N/A`. The one repository the server runs locally shows a real number.

| Repository | Runs on | Total size |
|---|---|---|
| oci-dev-1 | server | `475.96 MB` |
| oci-dev-2 | agent | `N/A` |
| oci-dev-3 | agent | `N/A` |
| home-em | agent | `N/A` |

All four are Borg 2 and healthy. Archive count, last backup and encryption all populate fine — only the size is missing.

## What I found

`_update_agent_repository_stats` in `app/api/repositories.py` takes the size from the rinfo result:

```python
stats = (rinfo.get("cache") or {}).get("stats") or {}
size_bytes = stats.get("unique_csize") or stats.get("unique_size")
```

That is the Borg 1 shape. Borg 2 does not return it. Here is what `repo-info --json` actually gives on borg 2.0.0b23:

```
top-level keys: ['cache', 'encryption', 'repository', 'security_dir']
cache keys:     ['path']
```

No `cache.stats`, so `size_bytes` is always `None` and `total_size` is never set.

The server-local repository is unaffected because `calculate_total_size_bytes` runs `du` on it. But that function handles only two cases — a repository with `host` set, or a local path. An agent repository is neither, so nothing runs for it.

The comment in the code says the size is unknowable for a remote Borg 2 repository. That is true of `repo-info`, but the agent is already running commands on the machine that holds the repository, so it can measure it there.

## One thing that confused me

The repository detail modal does show a size. That comes from `borg2 info --json` and is the newest archive's logical size, not the repository's size on disk. On the same repository I get 44.02 GiB in the modal and 42.61 GiB on disk. Both are labelled just "size", so it looks like a contradiction until you dig in.

## Suggested fix

Ask the agent to measure the repository, which is the same thing the server already does for its own.

I looked at `compact --stats` first since it reports pack size, but borg refuses `--stats` together with `--dry-run`, and a real `compact` rewrites the repository — not something that should happen on a stats refresh.

`du` on the agent works, and is cheap: about 9 ms on my 42.6 GiB repository (932 files), because Borg 2 packs into a few large files.

Four files need the new job kind. Missing any one of them fails differently, and two of those failures are quiet.

**1. Agent: build the command**

```diff
--- a/agent/borg_ui_agent/repository_ops.py
+++ b/agent/borg_ui_agent/repository_ops.py
@@ -39,6 +39,7 @@
     "repository.prune",
     "repository.compact",
     "repository.rclone_sync",
+    "repository.disk_usage",
 }
 
 # Kill a streaming extract only when no bytes have flowed for this long — a
@@ -106,6 +107,11 @@
         return cmd
 
     def build_command(self, *, rclone_config_path: Optional[str] = None) -> list[str]:
+        if self.job_kind == "repository.disk_usage":
+            if not self.repository_path:
+                raise ValueError("repository.disk_usage requires a repository path")
+            return ["du", "-sb", "--", self.repository_path]
+
         if self.job_kind == "repository.rclone_sync":
             rclone = _rclone_operation(self.operation)
             remote_name = _require_non_empty_string(
@@ -500,6 +506,7 @@
         "repository.list_archives",
         "repository.delete_archive",
         "repository.break_lock",
+        "repository.disk_usage",
     }:
         try:
             return _execute_short_repository_operation(
```

**2. Agent: advertise it and route it**

```diff
--- a/agent/borg_ui_agent/runtime.py
+++ b/agent/borg_ui_agent/runtime.py
@@ -40,6 +40,7 @@
     "repository.prune",
     "repository.compact",
     "repository.rclone_sync",
+    "repository.disk_usage",
 ]
 
 JOB_HANDLERS = {
@@ -59,6 +60,7 @@
     "repository.prune": execute_repository_operation_job,
     "repository.compact": execute_repository_operation_job,
     "repository.rclone_sync": execute_repository_operation_job,
+    "repository.disk_usage": execute_repository_operation_job,
 }
```

**3. Server: let it be dispatched**

```diff
--- a/app/services/repository_executor.py
+++ b/app/services/repository_executor.py
@@ -35,6 +35,7 @@
     "repository.prune",
     "repository.compact",
     "repository.rclone_sync",
+    "repository.disk_usage",
 }
```

This one caught me out. Without it the server answers `400 backend.errors.agents.unsupportedJobKind` before the agent is ever asked. That message sounds like the agent is out of date, but the agent was advertising the capability correctly the whole time.

**4. Server: admission, classed as a read**

```diff
--- a/app/services/job_admission.py
+++ b/app/services/job_admission.py
@@ -36,6 +36,7 @@
 OPERATION_REPOSITORY_LIST_ARCHIVE_CONTENTS = "repository.list_archive_contents"
 OPERATION_REPOSITORY_EXTRACT_ARCHIVE_FILE = "repository.extract_archive_file"
 OPERATION_BREAK_LOCK = "break_lock"
+OPERATION_DISK_USAGE = "repository.disk_usage"
 # Sentinel for an active repository agent job whose kind we don't recognize.
 # Classed WRITE so admission fails closed -- an unknown job might hold a borg
 # lock, and break_lock must never run alongside it.
@@ -77,6 +78,8 @@
     OPERATION_REPOSITORY_LIST_ARCHIVES,
     OPERATION_REPOSITORY_LIST_ARCHIVE_CONTENTS,
     OPERATION_REPOSITORY_EXTRACT_ARCHIVE_FILE,
+    # du takes no repository lock, so it must not be classed as a write.
+    OPERATION_DISK_USAGE,
 }
 
 AGENT_JOB_KIND_OPERATIONS = {
@@ -93,6 +96,7 @@
     "repository.list_archive_contents": OPERATION_REPOSITORY_LIST_ARCHIVE_CONTENTS,
     "repository.extract_archive_file": OPERATION_REPOSITORY_EXTRACT_ARCHIVE_FILE,
     "repository.restore": OPERATION_RESTORE,
+    "repository.disk_usage": OPERATION_DISK_USAGE,
 }
```

The `READ_OPERATIONS` membership matters: anything absent from that set is treated as `repository_write`, and a write-classed size check blocks a backup queued behind it with `repositoryOperationActive`.

**5. Server: use it when rinfo left the size unset**

```diff
--- a/app/api/repositories.py
+++ b/app/api/repositories.py
@@ -876,6 +876,34 @@
                 repository=repository.name,
                 error=str(e),
             )
+
+        # Borg 2 exposes no size via repo-info, so ask the agent to measure
+        # the repository directory it already has locally. Only when rinfo
+        # left total_size unset, so Borg 1 repositories are unaffected.
+        if total_size is None:
+            try:
+                du_job = queue_agent_repository_operation_job(
+                    db, repository, job_kind="repository.disk_usage"
+                )
+                await dispatch_agent_job_best_effort(
+                    db, du_job, repository_id=repository.id
+                )
+                du_result = await wait_for_agent_repository_operation_job(
+                    db, du_job.id, timeout_seconds=timeouts["info_timeout"]
+                )
+                du_meta = du_result or {}
+                if du_meta.get("return_code", 0) == 0:
+                    # `du -sb` prints "<bytes>\t<path>".
+                    first = (du_meta.get("stdout") or "").strip().split("\n")[0]
+                    size_field = first.split()[0] if first.split() else ""
+                    if size_field.isdigit() and int(size_field) > 0:
+                        total_size = format_bytes(int(size_field))
+            except Exception as e:
+                logger.warning(
+                    "agent disk-usage for stats refresh failed",
+                    repository=repository.name,
+                    error=str(e),
+                )
 
         if list_ok:
             repository.archive_count = archive_count
```

The `total_size is None` guard leaves Borg 1 repositories on the existing path.

I am running all five locally and every repository now reports a size — the largest reads 42.61 GB, which matches `du -sb` on that machine.

One last thing worth mentioning: when this fails for a repository it logs a warning, but the refresh still reports `success=4 errors=0`. So it fails quietly and the only symptom is `N/A`.

## Environment

- borg-ui 2.2.6, agent 0.1.1
- borg 2.0.0b23
- Borg 2 repositories, agent-executed, chacha20-poly1305

dolifer:
## Description

Agent-executed Borg 2 repositories always show `Total size: N/A`.

`_update_agent_repository_stats` reads the size from `cache.stats.unique_csize`, which is a Borg 1 shape. Borg 2's `repo-info --json` returns only `cache.path`, `encryption`, `repository` and `security_dir`, so `size_bytes` is always `None`:

```
top-level keys: ['cache', 'encryption', 'repository', 'security_dir']
cache keys:     ['path']
```

Server-local repositories are unaffected, because `calculate_total_size_bytes` runs `du` on them — but that path handles only a repository with `host` set or a local path, and an agent repository is neither, so nothing measures it.

This adds a `repository.disk_usage` job kind that runs `du -sb` on the agent, which is already executing commands on the machine that holds the repository. It is guarded on `total_size` still being unset, so Borg 1 repositories keep using `cache.stats` exactly as before.

`compact --stats` also reports on-disk pack size, but borg refuses `--stats` together with `--dry-run`, and a real `compact` rewrites the repository — not something that should run on a scheduled stats refresh.

Two details worth calling out for review:

- The job kind is registered in **all four** places it has to be — `DEFAULT_CAPABILITIES` and `JOB_HANDLERS` in the agent, `AGENT_JOB_KIND_OPERATIONS` in admission, and `REPOSITORY_OPERATION_CAPABILITIES` in the executor. With the last one missing the server answers `400 unsupportedJobKind` before any agent is asked, which reads as an out-of-date agent even though the agent advertises the capability correctly.
- It is classed as a **read**. Anything absent from `READ_OPERATIONS` is treated as `repository_write`, and a write-classed size probe would block a backup queued behind it with `repositoryOperationActive`.

Cost is small: about 9 ms on a 42.6 GiB repository (932 files), since Borg 2 packs into a few large files.

## Related Issue

Fixes #874

## Type of Change

- [ ] 📖 Documentation update
- [ ] 🎨 UI/UX improvement
- [ ] ♻️ Refactoring

## Testing

Running these changes against four repositories — three agent-executed, one server-local. All four now report a size; the largest reads `42.61 GB`, matching `du -sb` on that machine. Before the change the three agent repositories showed `N/A` and the server-local one was correct.

- [x] All existing tests pass

`tests/unit/test_repository_disk_usage.py` covers the command built (`du -sb --`, bytes rather than a human-readable suffix), the empty-path guard, agent advertisement and handler registration, admission and dispatch registration, and the read classification. Each test was checked against a deliberately broken version of the code to confirm it fails.

```
tests/unit/test_repository_disk_usage.py .....                           [100%]
5 passed

tests/unit
2745 passed, 11 skipped
```

## Checklist

One note on documentation: I updated the comment above `cache.stats` that said the remote size is unknowable, since that is no longer true once the agent measures it. I did not find user-facing docs describing the size field, but happy to add something if there is a page it belongs on.

### DIFF
--- a/agent/borg_ui_agent/repository_ops.py
+++ b/agent/borg_ui_agent/repository_ops.py
@@ -40,6 +40,7 @@ REPOSITORY_JOB_KINDS = {
     "repository.prune",
     "repository.compact",
     "repository.rclone_sync",
+    "repository.disk_usage",
 }
 
 # Kill a streaming extract only when no bytes have flowed for this long — a
@@ -107,6 +108,13 @@ class RepositoryOperationPayload:
         return cmd
 
     def build_command(self, *, rclone_config_path: Optional[str] = None) -> list[str]:
+        if self.job_kind == "repository.disk_usage":
+            if not self.repository_path:
+                raise ValueError(
+                    "repository.disk_usage requires a repository path"
+                )
+            return ["du", "-sb", "--", self.repository_path]
+
         if self.job_kind == "repository.rclone_sync":
             rclone = _rclone_operation(self.operation)
             remote_name = _require_non_empty_string(
@@ -517,6 +525,7 @@ def execute_repository_operation_job(
         "repository.list_archives",
         "repository.delete_archive",
         "repository.break_lock",
+        "repository.disk_usage",
     }:
         try:
             return _execute_short_repository_operation(
--- a/agent/borg_ui_agent/runtime.py
+++ b/agent/borg_ui_agent/runtime.py
@@ -41,6 +41,7 @@ DEFAULT_CAPABILITIES = [
     "repository.prune",
     "repository.compact",
     "repository.rclone_sync",
+    "repository.disk_usage",
     "agent.list_scripts",
     "script.run",
 ]
@@ -63,6 +64,7 @@ JOB_HANDLERS = {
     "repository.prune": execute_repository_operation_job,
     "repository.compact": execute_repository_operation_job,
     "repository.rclone_sync": execute_repository_operation_job,
+    "repository.disk_usage": execute_repository_operation_job,
 }
 
 
--- a/app/api/repositories.py
+++ b/app/api/repositories.py
@@ -874,7 +874,8 @@ async def _update_agent_repository_stats(repository: Repository, db: Session) ->
             rinfo = json.loads((rinfo_result or {}).get("stdout") or "{}")
             encryption_mode = (rinfo.get("encryption") or {}).get("mode")
             # Borg 1 `info --json` exposes repo-level dedup size in cache.stats;
-            # Borg 2 `repo-info --json` does not (remote size is unknowable).
+            # Borg 2 `repo-info --json` does not. For Borg 2 the size comes
+            # from the agent instead, below.
             stats = (rinfo.get("cache") or {}).get("stats") or {}
             size_bytes = stats.get("unique_csize") or stats.get("unique_size")
             if isinstance(size_bytes, (int, float)) and size_bytes > 0:
@@ -885,6 +886,35 @@ async def _update_agent_repository_stats(repository: Repository, db: Session) ->
                 repository=repository.name,
                 error=str(e),
             )
+
+        # Borg 2 reports no size through repo-info, so ask the agent to
+        # measure the repository directory it already holds locally. Guarded
+        # on total_size so Borg 1 repositories keep using cache.stats above.
+        if total_size is None:
+            try:
+                du_job = queue_agent_repository_operation_job(
+                    db, repository, job_kind="repository.disk_usage"
+                )
+                await dispatch_agent_job_best_effort(
+                    db, du_job, repository_id=repository.id
+                )
+                du_result = await wait_for_agent_repository_operation_job(
+                    db, du_job.id, timeout_seconds=timeouts["info_timeout"]
+                )
+                du_meta = du_result or {}
+                if du_meta.get("return_code", 0) == 0:
+                    # `du -sb` prints "<bytes>\t<path>".
+                    stdout = (du_meta.get("stdout") or "").strip()
+                    first = stdout.split("\n")[0] if stdout else ""
+                    fields = first.split()
+                    if fields and fields[0].isdigit() and int(fields[0]) > 0:
+                        total_size = format_bytes(int(fields[0]))
+            except Exception as e:
+                logger.warning(
+                    "agent disk-usage for stats refresh failed",
+                    repository=repository.name,
+                    error=str(e),
+                )
 
         if list_ok:
             repository.archive_count = archive_count
--- a/app/services/job_admission.py
+++ b/app/services/job_admission.py
@@ -36,6 +36,7 @@ OPERATION_REPOSITORY_LIST_ARCHIVES = "repository.list_archives"
 OPERATION_REPOSITORY_LIST_ARCHIVE_CONTENTS = "repository.list_archive_contents"
 OPERATION_REPOSITORY_EXTRACT_ARCHIVE_FILE = "repository.extract_archive_file"
 OPERATION_BREAK_LOCK = "break_lock"
+OPERATION_DISK_USAGE = "repository.disk_usage"
 # Sentinel for an active repository agent job whose kind we don't recognize.
 # Classed WRITE so admission fails closed -- an unknown job might hold a borg
 # lock, and break_lock must never run alongside it.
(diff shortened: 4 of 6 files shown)


dolifer:
Good catch — confirmed and fixed in the second commit.

You're right that `ensure_repository_admission` rejects a requested write when *any* work is active, so classing disk usage as a plain read still let an in-flight size check refuse a backup. Since the stats refresh runs on a timer, that collision would have landed on scheduled backups at random intervals — worse than the bug I was fixing.

Rather than special-case the operation in the conflict matrix, I added a third class, since "observes without opening the repository" is a genuinely different thing from a read:

- `repository_read` opens the repository, so it holds a lock and must still yield to a write.
- `repository_observe` only stats the directory. It holds nothing and has nothing to wait for.

Observation work is skipped from both sides of the conflict check — it never blocks anything, and it can start while anything else is running. The duplicate-operation guard runs before that loop and is untouched, so two concurrent size checks on one repository are still rejected.

Added `test_backup_can_queue_while_disk_usage_is_active`, which queues an active `repository.disk_usage` agent job and asserts a backup is admitted. I checked it against the previous behaviour to be sure it actually catches this: with the observation skip removed it fails, so it would have caught the case you flagged.

`2746 passed, 11 skipped`, ruff clean.

PR #876, opened 2026-09-02
fix: allow cloud mirrors on managed-agent repositories

### LINKED ISSUE #867: [Bug]: Cloud mirror never runs on a managed-agent repository — repository.rclone_sync missing from AGENT_JOB_KIND_OPERATIONS
### Summary

Enabling a cloud mirror on a managed-agent repository writes the mirror row fine, then every sync dies in milliseconds:

```
Unsupported agent repository operation: repository.rclone_sync
```

It's rejected at admission and never dispatched — nothing reaches the agent.

`app/services/job_admission.py:82` maps agent job kinds to operations, and `repository.rclone_sync` isn't there, so `operation_for_agent_job_kind()` raises:

```python
def operation_for_agent_job_kind(job_kind: str) -> str:      # line 131
    try:
        return AGENT_JOB_KIND_OPERATIONS[job_kind]
    except KeyError as exc:
        raise ValueError(f"Unsupported agent repository operation: {job_kind}") from exc
```

### Steps to reproduce

1. Register a managed agent and create an agent-executed repository on it.
2. Enable a cloud mirror on that repository, policy `after_success`.
3. Trigger a sync.

### Actual behavior

```
job 20:  started 15:27:57.652  →  failed 15:27:57.658      (6 ms)
error:   Unsupported agent repository operation: repository.rclone_sync
```

`GET /api/repositories/<id>`:

```json
"rclone_storage": {
  "backend": "rclone",
  "sync_direction": "agent_to_remote",
  "sync_status": "failed",
  "last_sync_error": "Unsupported agent repository operation: repository.rclone_sync",
  "agent_machine_status": "online"
}
```

### Expected behavior

The sync is dispatched to the agent, which runs `rclone sync` and reports back.

### Borg UI version

2.2.6 (digest `sha256:e04c5a7f…`, also `latest`), agent 0.1.1. Same on `main` and `v2.3.0-alpha.1`.

### Deployment method

Docker Compose.

### Host platform

Server: Ubuntu 24.04 (arm64). Agents: Ubuntu 24.04 (arm64), Debian 12 (x86_64).

### BorgBackup version

borg 2.0.0b23.

### Additional context

One of my four repositories mirrors: the one Borg UI runs itself, where sync is a local subprocess with no agent job. The three agent-executed ones all fail the same way.

Adding the kind fixes it:

```python
OPERATION_RCLONE_SYNC = "repository.rclone_sync"
"repository.rclone_sync": OPERATION_RCLONE_SYNC,
```

I put it in `WRITE_OPERATIONS` so a sync can't overlap a prune or compact, but you'll know better. Happy to open a PR.

With that change the sync runs to completion — 42.6 GiB across 927 files, matching on both sides.

dolifer:
## Description

Enabling a cloud mirror on an agent-executed repository saves fine, then every sync fails within milliseconds:

```
Unsupported agent repository operation: repository.rclone_sync
```

`AGENT_JOB_KIND_OPERATIONS` in `job_admission.py` has no entry for `repository.rclone_sync`, so `operation_for_agent_job_kind()` raises and the job is rejected at admission. Nothing reaches the agent.

Everything else for the feature is already in place, which is what makes this look like an oversight rather than an unfinished feature — the executor allows the kind for dispatch, and both agent modules handle it.

The operation also needs classifying. `operation_class_for()` raises on an operation in neither `READ_OPERATIONS` nor `WRITE_OPERATIONS`, so adding the mapping on its own just moves the failure one step later. rclone reads the repository and writes only to the remote, so it sits with the reads — the same shape as `restore`, which also reads the repository and writes elsewhere.

## Related Issue

Fixes #867

## Type of Change

- [ ] 📖 Documentation update
- [ ] 🎨 UI/UX improvement
- [ ] ♻️ Refactoring

## Testing

Running this against three agent-executed repositories mirroring to Google Drive. Before the change every sync was rejected at admission; after it the mirrors run and complete.

- [x] All existing tests pass

The test covers both halves — the mapping and the classification — because either one missing produces a different failure. Both were checked against a deliberately broken version of the code to confirm the test fails.

```
tests/unit/test_job_admission.py ....                                    [100%]
4 passed

tests/unit
2741 passed, 11 skipped
```

## Checklist

### DIFF
--- a/app/services/job_admission.py
+++ b/app/services/job_admission.py
@@ -37,6 +37,7 @@ OPERATION_REPOSITORY_LIST_ARCHIVE_CONTENTS = "repository.list_archive_contents"
 OPERATION_REPOSITORY_EXTRACT_ARCHIVE_FILE = "repository.extract_archive_file"
 OPERATION_BREAK_LOCK = "break_lock"
 OPERATION_DISK_USAGE = "repository.disk_usage"
+OPERATION_RCLONE_SYNC = "repository.rclone_sync"
 # Sentinel for an active repository agent job whose kind we don't recognize.
 # Classed WRITE so admission fails closed -- an unknown job might hold a borg
 # lock, and break_lock must never run alongside it.
@@ -83,6 +84,9 @@ READ_OPERATIONS = {
     OPERATION_REPOSITORY_LIST_ARCHIVES,
     OPERATION_REPOSITORY_LIST_ARCHIVE_CONTENTS,
     OPERATION_REPOSITORY_EXTRACT_ARCHIVE_FILE,
+    # rclone reads the repository and writes only to the remote, so it
+    # cannot corrupt local state the way a write operation can.
+    OPERATION_RCLONE_SYNC,
 }
 
 # du stats the repository directory; it never opens the repository, so it
@@ -106,6 +110,7 @@ AGENT_JOB_KIND_OPERATIONS = {
     "repository.extract_archive_file": OPERATION_REPOSITORY_EXTRACT_ARCHIVE_FILE,
     "repository.restore": OPERATION_RESTORE,
     "repository.disk_usage": OPERATION_DISK_USAGE,
+    "repository.rclone_sync": OPERATION_RCLONE_SYNC,
 }
 
 MAINTENANCE_MODEL_OPERATIONS = {
--- a/tests/unit/test_job_admission.py
+++ b/tests/unit/test_job_admission.py
@@ -115,7 +115,9 @@ def test_rclone_sync_is_mapped_and_classed_as_a_read():
     # "Unsupported agent repository operation: repository.rclone_sync",
     # never reaching the agent that supports it.
     assert AGENT_JOB_KIND_OPERATIONS["repository.rclone_sync"] == OPERATION_RCLONE_SYNC
-    assert operation_for_agent_job_kind("repository.rclone_sync") == OPERATION_RCLONE_SYNC
+    assert (
+        operation_for_agent_job_kind("repository.rclone_sync") == OPERATION_RCLONE_SYNC
+    )
     # rclone reads the repository and writes only to the remote. It also has
     # to be in one of the two sets: operation_class_for() raises on an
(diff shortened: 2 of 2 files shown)


dolifer:
Thanks for looking — I don't think this one applies. `operation_class_for` is already imported at module level in that file, above the tests:

```python
from app.services.job_admission import (
    OPERATION_BACKUP,
    OPERATION_BREAK_LOCK,
    OPERATION_CLASS_REPOSITORY_WRITE,
    ensure_repository_admission,
    operation_class_for,      # <- line 11, pre-existing
)
```

The local import inside my test only pulls in the names that *aren't* already there (`AGENT_JOB_KIND_OPERATIONS`, `OPERATION_CLASS_REPOSITORY_READ`, `OPERATION_RCLONE_SYNC`, `operation_for_agent_job_kind`), so the assertion resolves fine.

Verified rather than assumed:

```
tests/unit/test_job_admission.py::test_rclone_sync_is_mapped_and_classed_as_a_read
1 passed

tests/unit/test_job_admission.py
4 passed
```

No `NameError`. Happy to move the whole block up to the module-level import for consistency if you'd prefer that style — it reads a bit better either way — but it isn't a correctness issue.

dolifer:
I've moved the imports to the module-level block, so this is resolved either way.

To be clear on the original finding though: there was no `NameError`. `operation_class_for` was already imported at line 11, about a hundred lines above the test, and the function-local block only added the names that weren't already there. The test passed both before and after:

```
tests/unit/test_job_admission.py
4 passed
```

But the local import block is what made it look ambiguous, so hoisting it costs nothing and removes the question. Same names, one block.

Re-checked the mutations after the change to make sure the test still guards what it should — removing the mapping fails it, and removing the read classification fails it. `2741 passed, 11 skipped`, ruff clean.

karanhudia:
@dolifer Thanks for the quick fix. 🔥

dolifer:
@karanhudia you're welcome, just fixing the things that bites me during setup of homelab 🤣

PR #878, opened 2026-09-02
fix: send an agent real rclone credentials, not the redacted ones

### LINKED ISSUE #868: [Bug]: Agent cloud mirror gets redacted credentials — _agent_rclone_config reads redacted_config
### Summary

The rclone config sent to an agent for `repository.rclone_sync` is built from `redacted_config`, so `token`, `client_id` and `client_secret` arrive as the literal string `***`. The agent writes them into its temp rclone.conf and rclone fails before touching the network.

`app/services/rclone_repository_service.py:300`:

```python
def _agent_rclone_config(self, remote: RcloneRemote) -> dict[str, str]:
    values = dict(remote.redacted_config or {})
```

### Steps to reproduce

1. Enable a cloud mirror on an agent-executed repository.
2. Trigger a sync.

### Actual behavior

The job reaches the agent and exits 1 immediately:

```
repository.rclone_sync exited with code 1
```

rclone's own error, reproduced by hand with the same config the agent writes:

```
Failed to create file system for "GDrive:borg": drive: failed when making
oauth client: failed to create oauth client: invalid character '*' looking
for beginning of value
```

The captured job payload shows the values that arrived:

```
config KEYS: ['client_id', 'client_secret', 'scope', 'token', 'type']
  type = drive
  scope = drive
  token         = <len 3>      # "***"
  client_id     = <len 3>
  client_secret = <len 3>
```

### Expected behavior

The agent receives credentials it can authenticate with, and the sync runs.

### Borg UI version

2.2.6, agent 0.1.1. Same code on `main`.

### Deployment method

Docker Compose.

### Host platform

Server: Ubuntu 24.04 (arm64). Agents: Ubuntu 24.04 (arm64), Debian 12 (x86_64).

### BorgBackup version

borg 2.0.0b23.

### Additional context

`RcloneRemote` only stores `redacted_config` — the real values live in the rclone.conf the server manages — so this isn't a one-word fix, and it's worth deciding alongside how secrets reach agents generally. Backup payloads already send `BORG_PASSPHRASE` as `{"value": ...}`, and the managed-agent spec's `{"secret_ref": ...}` suggests you had a different direction in mind.

I worked around it locally by putting the real credentials in the agent's own environment (`RCLONE_CONFIG_<REMOTE>_<KEY>`), which override the config file, so nothing extra travels in the payload.

Note this only surfaces once #867 is fixed — before that the job never reaches the agent at all.

dolifer:
## Description

The rclone config sent to an agent for `repository.rclone_sync` is built from `remote.redacted_config`, so `token`, `client_id` and `client_secret` arrive as the literal string `***`. The agent writes them into its own rclone.conf and rclone fails before touching the network:

```
invalid character '*' looking for beginning of value
```

which surfaces only as `repository.rclone_sync exited with code 1`.

`redacted_config` exists for display, and it is the only config the `RcloneRemote` row carries — the real values live in the server's rclone.conf. This reads the remote's section from there instead.

If that file is missing or unreadable, the previous behaviour is kept as a fallback, so a misconfigured path still fails inside rclone with a message about the remote rather than sending a section with no `type` at all.

Worth noting this only shows up after #867 is fixed, since without that mapping the mirror never gets far enough to fail this way.

## Related Issue

Fixes #868

## Type of Change

- [ ] 📖 Documentation update
- [ ] 🎨 UI/UX improvement
- [ ] ♻️ Refactoring

## Testing

Running this against three agent-executed repositories mirroring to Google Drive. My own workaround until now was giving each agent its own rclone credentials through the environment, which is what made the cause visible — `RCLONE_CONFIG_<REMOTE>_<KEY>` overrides the config file, so the mirrors worked while the config the server sent was still unusable.

- [x] All existing tests pass

The tests cover real values reaching the agent, `type` being preserved, and the unreadable-file fallback. Checked against the previous implementation to confirm the first one fails when the config is built from `redacted_config`.

```
tests/unit/test_agent_rclone_config.py ...                               [100%]
3 passed

tests/unit
2743 passed, 11 skipped
```

## Checklist

One thing I'd like your view on: this sends real credentials to the agent, which is what makes the mirror work, but you may prefer a different shape — obscured values, a scoped token per agent, or having the agent hold its own remote config and the server send only the remote name. Happy to rework it if one of those fits the direction better.

### DIFF
--- a/app/services/rclone_repository_service.py
+++ b/app/services/rclone_repository_service.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import configparser
 import os
 import shlex
 import shutil
@@ -297,8 +298,38 @@ class RcloneRepositoryService:
             raise ValueError("SSH cloud mirror mount did not return a mount point")
         return mount_point, mount_id, mount_service
 
+    def _rclone_config_section(self, remote: RcloneRemote) -> dict[str, str]:
+        """Read a remote's real values from the server's rclone.conf.
+
+        redacted_config stores secrets as "***" for display, so it cannot be
+        used to build a working config for an agent.
+        """
+        config_path = remote.config_path or str(
+            Path(settings.rclone_config_root) / "rclone.conf"
+        )
+        parser = configparser.RawConfigParser()
+        try:
+            with open(config_path, encoding="utf-8") as handle:
+                parser.read_file(handle)
+        except (OSError, configparser.Error):
+            return {}
+        if not parser.has_section(remote.name):
+            return {}
+        return dict(parser.items(remote.name))
+
     def _agent_rclone_config(self, remote: RcloneRemote) -> dict[str, str]:
-        values = dict(remote.redacted_config or {})
+        # The agent writes these into its own rclone.conf and runs rclone
+        # against it, so the values have to be the real ones. Building this
+        # from redacted_config sent "***" as token, client_id and
+        # client_secret, and rclone failed before touching the network with
+        # "invalid character '*' looking for beginning of value" -- which
+        # surfaces only as "exited with code 1".
+        values = self._rclone_config_section(remote)
+        if not values:
+            # No readable section: fall back to what is known rather than
+            # sending nothing, so a misconfigured path fails in rclone with
+            # a message about the remote instead of a missing type.
+            values = dict(remote.redacted_config or {})
         values["type"] = str(values.get("type") or remote.provider).strip()
         return {
             str(key): _stringify_config_value(value)
--- a/app/services/rclone_repository_service.py
+++ b/app/services/rclone_repository_service.py
@@ -311,7 +311,10 @@ class RcloneRepositoryService:
         try:
             with open(config_path, encoding="utf-8") as handle:
                 parser.read_file(handle)
-        except (OSError, configparser.Error):
+        # UnicodeError too: a config with non-UTF-8 bytes raises
+        # UnicodeDecodeError, which is a ValueError rather than an OSError and
+        # would otherwise escape and fail the whole stats refresh.
+        except (OSError, UnicodeError, configparser.Error):
             return {}
         if not parser.has_section(remote.name):
             return {}
--- a/tests/unit/test_agent_rclone_config.py
+++ b/tests/unit/test_agent_rclone_config.py
@@ -1,0 +1,68 @@
+import pytest
+
+from app.database.models import RcloneRemote
+from app.services.rclone_repository_service import rclone_repository_service
+
+CONFIG = """[gdrive]
+type = drive
+scope = drive
+token = {"access_token":"real-token"}
+client_id = real-client-id
+client_secret = real-client-secret
+"""
+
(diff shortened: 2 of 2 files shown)


dolifer:
Fair point — fixed.

You're right that `type` alone was too weak an assertion: an implementation that dropped `redacted_config` and set only the provider would have passed it, which is exactly the behaviour the test exists to rule out.

Added the representative fields, and checked the test actually catches that case now — with the fallback changed to `values = {}` (keeping only the provider type) it fails, where before it passed:

```
tests/unit/test_agent_rclone_config.py
3 passed

tests/unit
2743 passed, 11 skipped
```

dolifer:
Good call — added.

The `except (OSError, configparser.Error)` clause does already catch `PermissionError`, since it's an `OSError` subclass, but you're right that nothing proved it. And it's the more realistic failure of the two: rclone.conf holds credentials and is normally 0600, so a server running as another user hits a permission error rather than a missing path.

Two tests added, one per arm of that except clause:

- `test_agent_config_falls_back_when_the_file_cannot_be_read` — file exists, `open` raises `PermissionError`
- `test_agent_config_falls_back_when_the_file_is_malformed` — a hand-edited/truncated file, which raises `configparser.Error` rather than `OSError`

The original missing-file test is unchanged.

Checked that each one actually exercises its arm: narrowing the clause to `except configparser.Error` fails two tests (missing file and unreadable), and narrowing it to `except OSError` fails the malformed one.

```
tests/unit/test_agent_rclone_config.py
5 passed

tests/unit
2745 passed, 11 skipped
```

dolifer:
Confirmed and fixed — this one was a real bug, not just missing coverage.

`UnicodeDecodeError` inherits from `ValueError`, not `OSError`, so it walked straight past `except (OSError, configparser.Error)`. I reproduced it before changing anything: a config containing a `0xff` byte raises `UnicodeDecodeError` out of `_agent_rclone_config`, which would have propagated into the stats refresh rather than falling back.

```
E   UnicodeDecodeError: 'utf-8' codec can't decode byte 0xff in position 30: invalid start byte
```

Added `UnicodeError` to the handler (the parent class, so it also covers an encode error if the path ever changes) and a regression test with genuinely invalid bytes.

All three arms are now independently covered — dropping `OSError` fails two tests, dropping `UnicodeError` fails one, dropping `configparser.Error` fails one:

```
tests/unit/test_agent_rclone_config.py
6 passed

tests/unit
2746 passed, 11 skipped
```

karanhudia:
This is the right direction and fixes the actual problem. An agent needs the real rclone credentials to run the mirror.

One follow-up for a separate PR is to avoid keeping those credentials as plaintext in the durable `AgentJob.payload`. The better design would keep them as an encrypted job secret, materialize them only when delivering the job to the authenticated agent through polling or WebSocket, and remove them once the job reaches a terminal state.

I do not want to hold this functional fix for that larger hardening work. Please open a follow-up issue or PR for it and we can merge this one separately.

Issue #887, opened 2026-09-03
Archive names from {now} templates render in the creating process's zone


ioanalytica:
Follow-up from the #871 discussion (cosmetic, low priority).

Archive names built from `{now}` templates render in the creating process's local zone, while stored timestamps are UTC. Across mixed-zone fleets (or after the TZ=UTC-at-source hardening, if applied only to machine-parsed invocations) archive *names* therefore neither sort consistently with each other nor align with the UTC timestamps shown next to them.

Nothing is functionally wrong — worth either documenting as expected behavior or normalizing the template evaluation (e.g. an explicit `{utcnow}` recommendation in the docs and defaults). Filed mainly so the observation from the #871 review does not get lost.

Issue #872, opened 2026-09-02
Remote direct execution: capture the remote host's timezone for archive-time parsing


ioanalytica:
Since #871, naive borg archive timestamps are parsed in the zone of the machine that created the archive: agents report their IANA zone, and the server's local zone is the fallback. Remote direct execution is the one remaining creator of archives without a reporting channel — borg runs on the remote host over SSH, so archives carry that host's local wall clock, and they are currently interpreted in the server's zone. That is exact whenever both share a timezone and otherwise bounded by the offset difference, but it can be made exact.

Suggested follow-up:
- Detect the remote host's IANA zone when an SSH connection is validated (e.g. `timedatectl show -p Timezone --value`, falling back to `cat /etc/timezone` or the `/etc/localtime` symlink target — the same ladder the agent uses locally).
- Store it on the SSH connection record (nullable, refreshed on re-validation).
- Pass it as `timezone_name` to `parse_borg_archive_time` when parsing archive lists of repositories fed by that connection — the parser already takes the zone as a parameter, so no parsing changes are needed.

Old values self-heal on the next stats refresh, as with the main fix.

ioanalytica:
This can be closed as superseded by #889.

The original premise was incorrect: Borg 1 renders archive timestamps in the timezone of the process **listing** the archives, not the host that created them. #889 pins machine-parsed Borg invocations to `TZ=UTC` and parses their timestamps accordingly, including archives created through remote direct execution.

Integration tests cover creating an archive in one timezone and listing it in another. Capturing the remote creator’s timezone on the SSH connection is therefore unnecessary for archive-time parsing.

ioanalytica:
Closing as superseded by #889 (merged), see the comment above.

PR #889, opened 2026-09-03
fix: run machine-parsed borg invocations under TZ=UTC

### LINKED ISSUE #884: Run machine-parsed borg invocations under TZ=UTC everywhere
Follow-up from the #871 discussion (architecture recommendation).

Borg renders timestamps in the local zone of the invoking process. For machine-parsed invocations that means the parser has to know or guess the render zone — the whole class of bug #871 fixed. The server-side stats listing already sidesteps this by forcing `TZ=UTC` into the borg environment (`_repository_stats_borg_env`); the same trick generalized would eliminate zone-guessing from parsing entirely:

- run every borg invocation whose output is machine-parsed under `TZ=UTC`, on the server and in the agent alike;
- keep local-zone rendering only for output shown to humans verbatim;
- the agent-reported zone from #871 then remains as the compatibility fallback for agents that predate the change.

Measured while verifying #871: Borg 2 (2.0.0b22 and b23) already renders listing timestamps with an explicit UTC offset in `repo-list --json` (bare and `--format` fast path), so its values are self-describing either way — this hardening mainly concerns the Borg 1 paths, where `--json` timestamps are naive local (measured on 1.4.5). The rendering contracts are pinned by `tests/integration/test_borg_timestamp_rendering_integration.py` (added in #871).

ioanalytica:
## Description

Borg renders listing timestamps in the local zone of the invoking process — naive, with no offset, on Borg 1. #871 taught the parse boundary to carry the render zone; this PR pins the render zone itself, as discussed there: every borg invocation whose timestamps the server interprets now runs under `TZ=UTC`, on the server and in the agent alike. The zone-guessing then only remains as the compatibility fallback for agents that predate the change.

**Server.** The borg1/borg2 wrapper methods whose output is machine-parsed (`list_archives`, `info_archive`, borg2 `rinfo`/`info_repo`) pin `TZ=UTC` themselves, as does the raw `repo-list`/`repo-info` command channel behind `/repositories/{id}/archives` and `/repositories/{id}/info`. The caller-level `_repository_stats_borg_env` is gone — the guarantee now sits at the invocation, not at one call site.

**Agent.** The machine-parsed job kinds (`repository.list_archives`, `repository.info`, `repository.rinfo`, `repository.archive_info`) run under `TZ=UTC`, and the agent now reports `"UTC"` as its zone — the field's contract from #871 is "the zone borg renders listings in", and that zone is now pinned. The compatibility matrix needs no version gate:

| | old server | new server |
|---|---|---|
| **old agent** | renders machine zone, reports machine zone ✓ | same ✓ (the #871 fallback) |
| **new agent** | renders UTC, reports "UTC" ✓ | same ✓ |

**Display.** Listing/info responses re-render borg1's (now naive-UTC) timestamps with an explicit `+00:00` offset before they leave the API. Without that, `new Date("2026-09-03T12:00:00")` in the frontend reads the value as browser-local wall clock — which is also why this is a user-visible fix, not just hardening: today the archives table is only correct when the browser's zone happens to match the server's.

**Deliberate boundary — what stays in the process zone:**
- `create`: its parsed output is the resolved archive name only; pinning `TZ` would change what `{now}` templates render, which is #887's discussion, not this PR's.
- Archive-contents listings (browse): file mtimes are relayed for display, not interpreted; pinning them would shift the browse view without a normalization pass to justify it.
- Remote direct execution (SSH sources): tracked separately in #872 — forcing env over SSH interacts with the sudo handling from #844.

## Related Issue

Fixes #884

## Type of Change

- [ ] 📖 Documentation update
- [ ] 🎨 UI/UX improvement
- [x] ♻️ Refactoring

## Testing

- [x] All existing tests pass

New tests: the borg1/borg2 wrapper methods pin `TZ=UTC` (and `run_backup` does not — the #887 boundary), the agent runs machine-parsed kinds under `TZ=UTC` and leaves others in the machine zone, the agent reports `"UTC"` regardless of the machine zone, and `_normalize_archive_listing_times` adds the offset, applies an old agent's reported zone, and passes unparseable values through unchanged.

```
pytest tests/unit -q
2828 passed, 11 skipped in 465.95s (0:07:45)

All checks passed! / 503 files already formatted
```

The rendering contracts themselves stay pinned by `tests/integration/test_borg_timestamp_rendering_integration.py` from #871 (borg1 naive-local, borg2 offset-carrying) — this PR builds on exactly those.

## Checklist

Not applicable; backend/agent change.

## Additional Context

One consequence worth calling out: with the render zone pinned, `agent_machines.timezone` no longer describes the machine — a new agent always reports `"UTC"`. Nothing else consumes the field (it exists for listing parse provenance), but if you'd rather keep the machine zone observable alongside, a separate field would be the way; happy to adjust.

The legacy `/archives/list` raw-stdout passthrough now relays UTC-rendered borg1 timestamps as-is. The main archive endpoints normalize; that one has no dict to normalize without parsing and re-serializing the whole document, and its consumers are gone from the frontend — left unchanged deliberately.

### DIFF
--- a/agent/borg_ui_agent/borg.py
+++ b/agent/borg_ui_agent/borg.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import os
 import platform
 import re
 import shutil
@@ -8,7 +7,6 @@ import subprocess
 from dataclasses import dataclass
 from pathlib import Path
 from typing import Iterable, Optional
-from zoneinfo import ZoneInfo
 
 
 @dataclass(frozen=True)
@@ -27,44 +25,16 @@ class BorgBinary:
         }
 
 
-def detect_timezone() -> Optional[str]:
-    """The machine's IANA zone name, best effort.
-
-    Borg writes archive timestamps in this machine's local time with no
-    offset; reporting the zone lets the server interpret them correctly.
-    Only a name zoneinfo can resolve is reported - DST needs the zone, a
-    bare offset would misplace archives from the other half of the year.
-    """
-    candidates = [os.environ.get("TZ")]
-    try:
-        candidates.append(Path("/etc/timezone").read_text(encoding="utf-8").strip())
-    except OSError:
-        pass
-    try:
-        # /etc/localtime -> .../zoneinfo/<Area/City> on most distributions.
-        target = Path("/etc/localtime").resolve()
-        parts = target.parts
-        if "zoneinfo" in parts:
-            candidates.append("/".join(parts[parts.index("zoneinfo") + 1 :]))
-    except OSError:
-        pass
-    for name in candidates:
-        if not name:
-            continue
-        try:
-            ZoneInfo(name)
-        except Exception:
-            continue
-        return name
-    return None
-
-
 def detect_platform() -> dict:
     return {
         "hostname": platform.node(),
         "os": platform.system().lower() or "unknown",
         "arch": platform.machine() or "unknown",
-        "timezone": detect_timezone(),
+        # The zone borg renders machine-parsed output in, which this agent
+        # pins to TZ=UTC (repository_ops.MACHINE_PARSED_JOB_KINDS). The server
+        # interprets naive listing timestamps with this value; agents that
+        # predate the pin report their detected machine zone instead.
+        "timezone": "UTC",
     }
 
 
--- a/agent/borg_ui_agent/repository_ops.py
+++ b/agent/borg_ui_agent/repository_ops.py
@@ -43,6 +43,17 @@ REPOSITORY_JOB_KINDS = {
     "repository.disk_usage",
 }
 
+# Kinds whose stdout the server parses timestamps out of. These run under
+# TZ=UTC so borg1's naive rendering is UTC wall clock; borg2 renders an
+# explicit offset either way. Contents listings (browse) stay in the machine
+# zone - their mtimes are relayed for display, not interpreted.
+MACHINE_PARSED_JOB_KINDS = {
+    "repository.info",
+    "repository.rinfo",
+    "repository.archive_info",
+    "repository.list_archives",
+}
+
 # Borg 2.0.0b22 split repo-create's single --encryption value into the cipher,
 # where the key is stored, and the id hash. The server sends the combined mode
 # name it stores, so the agent translates it the same way the server does for
@@ -549,6 +560,11 @@ def execute_repository_operation_job(
         )
 
     env = build_borg_env(payload.environment)
+    if payload.job_kind in MACHINE_PARSED_JOB_KINDS:
+        # The server parses timestamps out of these outputs; pin the render
+        # zone so they come out UTC. Applied after the server-sent overrides:
+        # the reported machine timezone is "UTC" on the same contract.
+        env["TZ"] = "UTC"
     if payload.job_kind == "repository.init":
         # Repo creation must not touch the shared pack cache: borgstore
         # rejects an already-populated cache directory on create, and borg
--- a/app/api/archives.py
+++ b/app/api/archives.py
@@ -43,7 +43,7 @@ from app.utils.borg_env import (
 from app.utils.ssh_utils import (
     resolve_repo_ssh_key_file,  # noqa: F401
 )  # Backward-compatible patch target for tests
-from app.utils.datetime_utils import serialize_datetime
+from app.utils.datetime_utils import serialize_borg_archive_time, serialize_datetime
 
 logger = structlog.get_logger()
 router = APIRouter()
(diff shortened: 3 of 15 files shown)


karanhudia:
The TZ pin is the right call, and I checked the part I was most suspicious about first: `list_archives` and `info_archive` going from `env=exec_env if exec_env else None` to `env=exec_env` is safe, because `_execute_command` starts from `os.environ.copy()` and merges on top, and `borg2._run` goes through `_base_env`. No single key env is reaching a subprocess. `archives.py` hardcoding `timezone_name="UTC"` is fine too, that endpoint calls `borg.info_archive` directly and has no agent branch.

One gap, and I think it belongs in this PR rather than after it.

`app/services/repository_info_sync.py` is not in the diff, and line 75 is the last place that still guesses the render zone:

```python
newest = _newest_archive_time(
    archives, timezone_name=agent_timezone_for_repository(db, repository)
)
```

`agent_timezone_for_repository` returns `None` for a repository that is not agent executed, and `None` tells `parse_borg_archive_time` to read naive values in the server's local zone. Both server side callers now hand it output from wrappers this PR pins to UTC:

- `app/api/repositories.py:6214`, local borg2 info, via `_run_repository_command_with_retries` into `_run_repository_command` where you set `env["TZ"] = "UTC"`
- `app/api/v2/repositories.py:642`, straight after `borg2.info_repo(...)`, which you pin as well

So on any server whose local zone is not UTC, `last_backup` lands off by that offset. Your own `test_naive_times_resolve_through_the_given_zone` describes this exact class of bug from the other direction: "assuming UTC pushed last_backup into the future on non-UTC agents". Same mistake, sign flipped.

On why this is not a follow-up: before this PR that line was correct. A server side listing genuinely was rendered in the server's zone, so `None` was the right answer. This change is what makes it wrong. It is not a pre-existing wart that happens to sit nearby, it is a line the diff breaks, so merging without it leaves main in a state where the render zone and the parse zone disagree on one path. The description says the guarantee "now sits at the invocation, not at one call site", and this is the call site that did not get the memo.

Fair to say the practical hit is small. borg2 normally emits offset carrying timestamps and then the naive branch never runs at all. But the naive branch is the whole reason `timezone_name` is threaded to this call site, and #871 shipped a test for it, so it is clearly treated as reachable.

Roughly four lines, and the two `None` cases need to stay separate:

```python
from app.services.repository_executor import (
    agent_timezone_for_repository,
    is_agent_executor,
)

zone = agent_timezone_for_repository(db, repository)
if zone is None and not is_agent_executor(repository):
    # Server side listings render under TZ=UTC (borg2.info_repo / rinfo).
    zone = "UTC"
newest = _newest_archive_time(archives, timezone_name=zone)
```

`or "UTC"` reads nicer but is wrong here. It would also swallow the agent repo whose agent never reported a zone, and that case is still an unknown machine zone, not UTC.

Leaving this as a plain comment rather than inline since the file is not part of the diff.

ioanalytica:
Agreed on all points, and thanks for checking the env-merge chain first - that was the riskiest-looking part of the diff.

The repository_info_sync gap is exactly as you describe: that line was correct before this PR and this PR is what breaks it, so it is fixed here (ab70299f) rather than as a follow-up. The fix keeps the two None cases separate, as you sketched:

- agent-executed: agent_timezone_for_repository as before - reported zone, or the server-local fallback for agents that never reported one (the #871 semantics, unchanged);
- everything else: "UTC", because that is now the render zone of both server-side callers.

Two tests pin it: a naive server-side value parses as UTC under a TZ=Europe/Berlin host (fails on the previous code), and an agent repository keeps the reported-zone interpretation. You are right that borg2's offset-carrying timestamps make the practical window small - but the naive branch is the reason timezone_name exists at that call site, so it should tell the truth.

ioanalytica:
Heads-up before merging: #888's operations runner imports _repository_stats_borg_env from app/api/repositories (executors/index.py line 18), and this PR originally removed that helper as redundant - the two sides merge cleanly textually but the combination would ImportError at startup. Fixed in ee04755c: the helper is reinstated as a thin compat shim (the wrappers pin TZ=UTC themselves now, so the double-pin is harmless), and the stack is rebased onto current main so CI validates exactly the combination that would land. Verified the operations executors import and their test suites (40 tests) on the new tip. #891/#892 rebased along (29e4c527 / 318ebbae); merge order unchanged.

ioanalytica:
Rebased the stack onto phase 2 (#897) - one composition from the conflict worth flagging: the rewritten sync_archive_stats_from_info computes timezone_name = agent_timezone_for_repository(...) again, which is None for server-executed repositories - the same server-local read this thread settled. Entirely understandable how: phase 2 and this fix were in flight at the same time. The rebase carries the executor-branch provenance into the new pipeline, so archive_fields_from_listing/apply_listing and the columns-only fallback all consume the corrected zone now (agent: reported zone, or the #871 server-local fallback for zone-less agents; otherwise "UTC"). Your three phase-2 sync tests pass unchanged alongside the two provenance tests. Tips: #889=27a138ba, #891=78fa9226, #892=d4d6a127.

karanhudia:
Heads up: the `CLEAN` label here is stale. Against current `main` this conflicts in `app/services/repository_info_sync.py` (and its test), since #897 landed there first.

The resolution is not mechanical: #897 passes `agent_timezone_for_repository(...)` straight into its new `apply_listing` call, and that has to become this PR's `is_agent_executor(repository) ... else "UTC"` branch (plus the import). Taking main's side silently parses server-side archive times in the wrong zone once `TZ=UTC` is pinned.

ioanalytica:
Rebased onto current main (ce93ace7). The repository_info_sync.py conflict with #897 is resolved by composing this PR's provenance into #897's apply_listing: the render zone is is_agent_executor(repository) ? agent_timezone_for_repository(...) : "UTC" (not main's straight agent_timezone_for_repository, which would parse server-side times in the server's zone once TZ=UTC is pinned). New tip b1a414a9; #891/#892 rebased along (9683db2c / 99ffebc1).

PR #890, opened 2026-09-03
fix(notifications): notify when no valid source paths remain

### LINKED ISSUE #858: [Bug]: Notifications not working
### Summary

It fails to send out notifications.

### Steps to reproduce

1. Set up a notification(I used NTFY or email) through the admin interface
2. Test it, should work.
3. Then set up a plan to fail on purpose.
4. Trigger it, or wait for it to trigger.
5. Nothing gets sent.

### Actual behavior

It doesnt send a notification. And there is no error for it.

### Expected behavior

It should notify me upon fail.

### Borg UI version

2.2.6

### Deployment method

Docker run

### Host platform

Debain 13 running Docker 29.7.2

### BorgBackup version

1.4.4

### Logs and screenshots

```shell

```

### Additional context

It is literally because notification_service.send_backup_failure() is not getting called by "if not processed_source_paths:" at line 1995 of backup_services.py, causing no notification to be triggered. Its in the source code on github.

ioanalytica:
## Description

The reporter's diagnosis in #858 is exactly right: the `if not processed_source_paths:` early return in `BackupService.execute_backup` marks the job failed, commits, and returns — without ever calling `notification_service.send_backup_failure()`. It is the only failure path in the server-side runner with that gap: pre-hook failures notify inline, post-hook and borg failures fall through to the shared epilogue, and the outer exception handler notifies too. This one returns before any of those.

The path triggers when the processed source list comes back empty — in practice, when every source is an SSH URL and every mount fails (a dead source host), since local paths pass through unvalidated. That is precisely the "backup never even started" case failure notifications exist for, and it was the silent one.

The fix mirrors the pre-hook failure block: send the failure notification (best-effort, log on error) before the early return.

## Related Issue

Fixes #858

## Type of Change

- [ ] 📖 Documentation update
- [ ] 🎨 UI/UX improvement
- [ ] ♻️ Refactoring

## Testing

- [x] All existing tests pass

The regression test drives `execute_backup` into the empty-source-paths return and asserts the job fails with `failedPrepareSourcePaths` *and* that `send_backup_failure` went out with that error. Verified against the previous implementation: the test fails without the fix.

```
pytest tests/unit/test_backup_service_mocks.py -q
29 passed

pytest tests/unit -q
2819 passed, 11 skipped in 472.82s (0:07:52)

All checks passed! / 503 files already formatted
```

## Checklist

## Additional Context

The reporter tested on 2.2.6 with both NTFY and email — for anyone else reproducing: only backups whose sources are exclusively SSH URLs can hit this path; a plan with local paths fails later inside borg and notifies through the shared epilogue. Agent-executed jobs notify through their own path since #870.

### DIFF
--- a/app/services/backup_service.py
+++ b/app/services/backup_service.py
@@ -2006,6 +2006,19 @@ class BackupService:
                 mqtt_service.sync_state_with_db(
                     db, reason="backup failed: no valid source paths"
                 )
+
+                # This early return never reaches the shared failure epilogue,
+                # so the notification has to go out here.
+                try:
+                    await notification_service.send_backup_failure(
+                        db, repository, job.error_message, job_id, job_name
+                    )
+                except Exception as e:
+                    logger.warning(
+                        "Failed to send backup failure notification",
+                        error=str(e),
+                    )
+
                 return
             logger.info(
                 "Source paths prepared",
--- a/tests/unit/test_backup_service_mocks.py
+++ b/tests/unit/test_backup_service_mocks.py
@@ -1508,3 +1508,63 @@ async def test_execute_backup_rejects_observe_mode(
 
     assert job.status == "failed"
     assert job.error_message == json.dumps({"key": "backend.errors.borg.unknownError"})
+
+
+@pytest.mark.asyncio
+async def test_empty_source_paths_failure_sends_notification(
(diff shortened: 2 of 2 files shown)


karanhudia:
Thanks for the clean fix 💣

PR #896, opened 2026-09-03
fix: keep a malformed job_id from unwinding the session

dolifer:
Follow-up to the two notes on #877, both of which were explicitly offered as follow-ups rather than merge blockers.

## The cast

> `_job_id_for_dispatch` gets called straight from the `run_session` loop with nothing around it, so a malformed `job_id` now raises there and unwinds the whole session. Before this change the same `int()` lived at the top of `_handle_command`, where a bad value only killed that one daemon worker and the session kept going.

Correct, and it was a regression I introduced in #877 — moving the cast onto the session thread widened the blast radius from one command to the whole session. Fixed as suggested: the cast returns `None` on a bad value, so the id stays out of the registry and `_handle_command` still fails it exactly as it did before, in its own worker.

Verified both directions: with the guard, no message shape reaches the session thread as an exception (`"abc"`, `{}`, `[1]`, `None` all return `None`, while `42` and `"7"` still register). Without it, `job_id: "abc"` raises `ValueError` on the session thread — the failure described above.

## The duplication

> The worry is narrower: three copies of one explanation drift apart, and the copy someone happens to read is not always the copy that got updated. One home per fact, with the other sites pointing at it, would age better.

Agreed, and the register-before-start cluster was three copies I added in one commit. Consolidated to one home:

| Site | Before | After |
|---|---|---|
| `run_session` block | 17 lines | 5 |
| `_job_id_for_dispatch` docstring | 8 lines | 8 — unchanged |
| race test docstring | 20 lines | 5 |

The docstring keeps the full explanation, since you called out the lockstep with `_handle_command`'s early returns as the thing a future change can quietly break. The other two point at it. Longest block in `session.py` is now 8 lines, against the 7–9 line norm on `main`.

I left the `STALE_AGENT_JOB_REQUEUE_AFTER` cluster alone — you named that invariant as keep-at-full-length, and two of its three copies predate #877, so trimming them here would pull prose churn from already-merged commits into this diff. Happy to take it separately if you want it.

## Verification

- `ruff check app tests` / `ruff format --check app tests` — clean
- `ruff check` / `format --check` on `session.py` — clean
- Full unit suite: 2831 passed, 11 skipped
- Mutation-checked both changes: removing the `try/except` reproduces the session-thread `ValueError`; reverting register-before-start turns the race test red, so the trimmed docstring still guards the behaviour

Net −18 lines.

### DIFF
--- a/agent/borg_ui_agent/session.py
+++ b/agent/borg_ui_agent/session.py
@@ -539,9 +539,11 @@ class AgentSessionRuntime:
         # _handle_command -- a daemon worker -- the same bad value only kills
         # that one command. Returning None keeps it out of the registry and
         # leaves _handle_command to fail it exactly as it did before.
+        # OverflowError is not hypothetical: json.loads turns 1e400 into
+        # float("inf"), and int(inf) raises it rather than ValueError.
         try:
             job_id = int(raw_job_id) if raw_job_id is not None else None
-        except (TypeError, ValueError):
+        except (TypeError, ValueError, OverflowError):
             return None
         if job_id is None or get_job_handler(command) is None:
             return None
--- a/agent/borg_ui_agent/session.py
+++ b/agent/borg_ui_agent/session.py
@@ -372,23 +372,11 @@ class AgentSessionRuntime:
                     continue
                 message = json.loads(raw_message)
                 if isinstance(message, dict) and message.get("type") == "command":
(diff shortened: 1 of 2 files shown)


karanhudia:
Looks good 💯 I think this works.

PR #898, opened 2026-09-04
fix(restore-check): address Borg 2 series archives by aid: selector

### LINKED ISSUE #863: [Bug]: Restore checks fail on every Borg 2 repository — archive selected by series name, which cannot resolve to one archive
## Summary

Scheduled and manual restore checks fail on **any Borg 2 repository that has more than one archive in a series**, with `repository.restore exited with code 4`.

The cause is that the restore check identifies the archive by **name**. For Borg 2 repositories Borg UI intentionally builds a *stable series name* (no timestamp), so every archive in a series shares one name and is distinguished by its **id**. `borg extract` requires the target to resolve to exactly one archive, so it refuses.

Backups, integrity checks, prune and compact are unaffected. Restores work correctly when performed by archive id, so this is a verification-path bug rather than data loss — but the dashboard reports `RESTORE: Failed` indefinitely, which is indistinguishable from a genuinely unrestorable repository.

## Environment

- Borg UI **2.2.6** (also present in `v2.3.0-alpha.1` and `main`)
- Borg **2.0.0b23**, repository created with `--encryption=chacha20-poly1305`
- Reproduced with both server-executed and managed-agent repositories
- Restore check mode: canary (also fails in `paths` mode — same code path)

## Steps to reproduce

1. Create a **Borg 2** repository.
2. Create a backup plan targeting it and leave `archive_name_template` at its default.
3. Run the plan **twice**, so the series contains two archives. Both are named `<plan_name>-daily`:

   ```
   borg2 -r <repo> repo-list
   ab12cd34  ...  myplan-daily  ...
   ef56gh78  ...  myplan-daily  ...
   ```

4. Enable a restore check on the repository and run it.

**Expected:** the check extracts from the latest archive and passes.

**Actual:** it fails immediately:

```
Restore verification failed on the agent: repository.restore exited with code 4
```

with Borg reporting:

```
Command error: ['myplan-daily'] needed to match precisely one archive, but matched 2.
```

## Minimal confirmation

The difference is only how the archive is addressed:

```console
$ borg2 -r <repo> extract --dry-run myplan-daily
Command error: ['myplan-daily'] needed to match precisely one archive, but matched 2.

$ borg2 -r <repo> extract --dry-run aid:ab12cd34
$ echo $?
0
```

## Cause

`app/services/backup_plan_execution_service.py` deliberately omits the timestamp for Borg 2:

```python
stable_series=getattr(repo, "borg_version", 1) == 2,
```

which `app/utils/archive_names.py` honours:

```python
if not stable_series:
    parts.append(timestamp)
```

That is correct Borg 2 behaviour. But `app/services/restore_check_service.py` then selects the archive by name, preferring `name` over `id`:

```python
def _get_archive_name(archive: dict | str | None) -> str:
    ...
    for key in ("name", "archive", "id"):
        value = archive.get(key)
        if isinstance(value, str) and value:
            return value
    return ""
```

The archive dict already carries a unique `id`; only the name is ambiguous.

## Suggested fix

For Borg 2 repositories, address the archive by id — `aid:<id>` — rather than by series name. The value is already present in the payload the check reads, so this need not change how archives are named or listed.

ioanalytica:
## Description

Exactly the failure @dolifer diagnosed in #863: the restore check identifies its target archive by **name**, and for Borg 2 repositories the UI deliberately builds a *stable series name* — so every archive in the series shares one name and `borg extract` refuses to resolve it once the series has more than one archive (`repository.restore exited with code 4`). Every Borg 2 repository therefore reports `RESTORE: Failed` from the second backup on, indistinguishable from a genuinely unrestorable repository.

The fix separates *display* from *addressing*, using the same `aid:` pattern the archive download path already uses (`_archive_extract_selector`): a new `_get_archive_selector` returns `aid:<id>` for Borg 2 archives (falling back to the name when no id is present) and the unique name as-is for Borg 1. Both restore paths address by selector now — the server-side `build_restore_extract_command` and the agent operation payload (`repository.restore` passes the archive string verbatim into `borg extract`, so `aid:` flows through with no agent change) — while `job.archive_name`, logs and notifications keep the human-readable series name.

## Related Issue

Fixes #863

## Type of Change

- [ ] 📖 Documentation update
- [ ] 🎨 UI/UX improvement
- [ ] ♻️ Refactoring

## Testing

- [x] All existing tests pass

New: selector unit tests (Borg 2 dict → `aid:`, no-id fallback, Borg 1 name, string pass-through) and an execute-path test with a two-archive series asserting the extract is addressed as `aid:<latest-id>` — never the series name — while the job row keeps the display name. Verified against the previous implementation: 5 tests fail without the fix.

```
pytest tests/unit/test_restore_check_service.py -q
11 passed

pytest tests/unit -q
2954 passed, 11 skipped in 490.36s (0:08:10)

ruff check app tests && ruff format --check app tests — clean
```

## Checklist

## Additional Context

CodeRabbit pre-review ran on the fork (ioanalytica#44): clean first pass, no findings. The latest-archive *selection* itself (`_select_latest_archive`, string max over rendered timestamps) is unchanged — within a single listing all values share one render zone, so the ordering is stable; the addressing was the broken half. Independent of the open timezone PRs (#889/#891/#892) and of the operations-runner work — no shared files.

### DIFF
--- a/app/services/restore_check_service.py
+++ b/app/services/restore_check_service.py
@@ -87,6 +87,24 @@ def _get_archive_name(archive: dict | str | None) -> str:
     return ""
 
 
+def _get_archive_selector(archive: dict | str | None, repository) -> str:
+    """Selector that resolves to exactly one archive.
+
+    Borg 2 archives in a series share one name and are told apart by id, so
+    extract needs the aid: selector - a bare series name matches every
+    archive in the series and borg refuses to pick one. Borg 1 names are
+    unique and are passed as-is.
+    """
+    name = _get_archive_name(archive)
+    if getattr(repository, "borg_version", 1) != 2:
+        return name
+    if isinstance(archive, dict):
+        archive_id = archive.get("id")
+        if isinstance(archive_id, str) and archive_id:
+            return f"aid:{archive_id}"
+    return name
+
+
 # Upper bounds so a delegated restore-check cannot wait on the agent forever:
 # fail if the job is never claimed (agent offline) or goes silent (agent died).
 _AGENT_OP_CLAIM_TIMEOUT_SECONDS = 120
@@ -238,6 +256,7 @@ class RestoreCheckService:
             archives = await BorgRouter(repository).list_archives(env=env)
             archive = _select_latest_archive(archives)
             archive_name = _get_archive_name(archive)
+            archive_selector = _get_archive_selector(archive, repository)
             if not archive_name:
                 job.status = "needs_backup" if use_canary else "failed"
                 job.error_message = (
@@ -308,7 +327,7 @@ class RestoreCheckService:
             async def run_extract(paths: list[str]) -> int | None:
                 cmd = BorgRouter(repository).build_restore_extract_command(
                     repository_path=repository.path,
-                    archive_name=archive_name,
+                    archive_name=archive_selector,
                     paths=paths,
                     remote_path=effective_repository_remote_path(repository),
                     bypass_lock=repository.bypass_lock,
@@ -520,6 +539,7 @@ class RestoreCheckService:
 
         archive = _select_latest_archive(archives)
         archive_name = _get_archive_name(archive)
+        archive_selector = _get_archive_selector(archive, repository)
         if not archive_name:
             job.status = "needs_backup" if use_canary else "failed"
             job.error_message = (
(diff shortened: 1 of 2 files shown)


ioanalytica:
@coderabbitai review

PR #900, opened 2026-09-04
fix(retention): take job log files along when purging rows

### LINKED ISSUE #895: Clean up retained job log files for all applicable job models
## Summary

`app/services/job_history_retention.py` deletes expired job rows through `purge_job_rows` and `_delete_chunked`. The row deletion does not remove the associated `log_file_path` file for applicable models in `_JOB_TABLES`.

This behavior predates the `Operation` model. Do not add an `Operation`-only fix.

## Required changes

Update the retention cleanup so it identifies and unlinks `log_file_path` files for every applicable job model in `_JOB_TABLES` when it deletes expired rows.

Preserve the existing safety ordering used by `purge_jobs_for_pruned_archives`:

1. Capture file paths before row deletion.
2. Commit the database deletion.
3. Unlink the captured files after the commit.
4. Tolerate missing files and filesystem unlink errors.

## Rationale

Without file cleanup, retention can delete database rows while their operation or job log files remain on disk. A special case for `Operation` would make retention behavior inconsistent across the existing job tables.

## Affected areas

- `app/services/job_history_retention.py`
  - `_JOB_TABLES`
  - `purge_job_rows`
  - `_delete_chunked`
  - `purge_jobs_for_pruned_archives` as the existing reference behavior
- Retention unit tests

## Acceptance criteria

- The retention path unlinks `log_file_path` files for all applicable job models deleted by `purge_job_rows`.
- The cleanup does not special-case `Operation`.
- A failed database commit does not remove files for rows that remain in the database.
- Missing log files and unlink failures do not stop the retention pass.
- Regression tests cover at least multiple applicable job models and confirm that unrelated rows and files remain intact.

## Backlinks

- PR: https://github.com/karanhudia/borg-ui/pull/888
- Review comment: https://github.com/karanhudia/borg-ui/pull/888#discussion_r3924209392

Requested by: @karanhudia

ioanalytica:
## Description

As specified in #895: `purge_job_rows` deleted expired job rows without removing the `log_file_path` files they named, so retention could clear the database while the log files lived on indefinitely. Only `Operation` rows had file cleanup — exactly the special case the issue asks to avoid.

The fix moves the behavior into `_delete_chunked` itself, uniformly for every model in `_JOB_TABLES` that has a `log_file_path` column: each chunk captures its rows' paths before the delete, commits the deletion, and only then unlinks — so a failed commit never leaves a surviving row pointing at a vanished file — with `missing_ok` and `OSError` tolerance. That is the same safety ordering `purge_jobs_for_pruned_archives` already uses. The `Operation` pre-clear in `purge_job_rows` is gone (the generic path covers it); `purge_operation_log_files` itself stays, because the log-retention window still uses it to drop files while keeping rows.

**Existing leftovers.** Installations have been accumulating orphaned files since the gap existed - rows long purged, files still on disk, and with the rows gone no DB-driven pass can ever find them again. A new `sweep_orphaned_log_files` retention phase reclaims them: `*.log` inside the log directory, file mtime older than the row-retention cutoff, and never a file some live `log_file_path` row still names. Deliberately a recurring retention phase rather than a one-shot migration - deleting filesystem files from an Alembic migration would be the wrong layer (side effects, no rollback), and the sweep also self-heals the one residual leak the fix's own ordering allows (a crash between a chunk's commit and its unlink).

## Related Issue

Fixes #895

## Type of Change

- [ ] 📖 Documentation update
- [ ] 🎨 UI/UX improvement
- [ ] ♻️ Refactoring

## Testing

- [x] All existing tests pass

New tests: (1) `run_retention` over expired rows across three models (CheckJob, CompactJob, RestoreCheckJob) plus one with an already-missing file and one inside the window - expired files are unlinked, the missing file does not break the purge, the fresh row keeps row and file; fails without the fix. (2) The orphan sweep removes an aged unreferenced file while keeping a referenced old file, a young orphan, and a non-.log file.

```
pytest tests/unit/test_job_history_retention.py -q
16 passed

pytest tests/unit -q
2950 passed, 11 skipped in 487.64s (0:08:07) - plus the sweep test added after that run; the retention suite reruns green (16 passed)

ruff check app tests && ruff format --check app tests — clean
```

## Checklist

## Additional Context

Independent of the open timezone stack (#889/#891/#892) and of #898 — no shared files; rebased onto current main (#897) before submission. CodeRabbit pre-review ran on the fork (ioanalytica#45): one round (shared log_file_path between an expired and a surviving row — fixed with a post-commit reference check plus regression test), final pass clean.

### DIFF
--- a/app/services/job_history_retention.py
+++ b/app/services/job_history_retention.py
@@ -37,8 +37,10 @@ the file; the manual cleanup endpoint runs VACUUM for that.
 from __future__ import annotations
 
 import re
-from datetime import timedelta
+from datetime import timedelta, timezone
 from pathlib import Path
+
+from app.config import settings as app_config
 from typing import Dict, Iterable, Optional
 
 import structlog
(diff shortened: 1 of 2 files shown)


ioanalytica:
@coderabbitai review

PR #946, opened 2026-09-07
feat(install): native install for Debian and Ubuntu hosts (bare metal, VM, LXC)

### LINKED ISSUE #941: Feature request: support additional installation methods (bare metal, VM, LXC)
## Summary

Currently Docker is the primary supported installation path. Add first-class support (and documented, step-by-step instructions) for installing Borg UI on additional platforms to widen distribution and make it easier to adopt in more environments.

## Motivation

Not everyone runs Docker. Supporting more installation methods lowers the barrier to entry and lets users pick whatever fits their existing infrastructure.

## Proposed installation methods

- **Bare metal** — install directly onto a Linux host, no container layer
- **KVM / VM** — run inside a virtual machine on KVM, Proxmox, or any hypervisor
- **LXC** — lightweight system container install, minimal overhead

## Notes

- Docker should remain the recommended/fastest path.
- Instructions for each method should live in the wiki so they can be kept current with releases.
- This is a feature request for discussion/scoping — not an implementation plan yet.

karanhudia:
Closes #941.

Docker was the only supported way to run Borg UI. This adds a native install path: bare metal, a KVM/Proxmox VM and an LXC container are all the same thing to the installer, a Debian or Ubuntu host with systemd. Docker stays the recommended path.

## What is here

| File | Role |
| --- | --- |
| `scripts/install.sh` | The installer. Provisions the host and can be re-run to upgrade. |
| `scripts/build-native-tarball.sh` | Builds the release artifact the installer downloads. |
| `.github/workflows/release.yml` | Publishes that artifact with a SHA-256 beside it. |
| `packaging/native/borg-ui.service` | systemd unit template. |
| `packaging/native/start.sh` | The container entrypoint's launch sequence, natively. |
| `tests/unit/test_native_install_packaging.py` | Guards the seams shell cannot cover. |

```bash
curl -fsSL https://raw.githubusercontent.com/karanhudia/borg-ui/main/scripts/install.sh | sudo bash
```

## Decisions worth reviewing

**No version constants in the installer.** Every Borg and rclone pin travels in the release tarball, generated at build time from `docker/runtime-base.env` and `Dockerfile.runtime-base`. A native host therefore runs exactly the Borg the Docker image runs, and cannot drift from it. Borg 1 comes from the same pinned static-binary manifest the server already serves to enrolling agents (`app/api/borg_binaries.json`), so no compiler is needed and every download is checksum-verified.

**Borg 2 is honest about what it cannot do.** Its static binary needs glibc 2.43, which no shipping release has (trixie is 2.41), so the realistic path is a source build, which needs OpenSSL 3.2 or newer because 2.0.0b24 takes argon2 from OpenSSL. On Debian 12 that is unavailable. The installer says so plainly and continues; Borg 1 repositories are unaffected.

**The service runs as root by default.** A host backup tool has to read files it does not own, so an unprivileged default would silently skip most of the host. `--service-user borg` takes the other trade, and the docs spell out what it costs. The systemd unit also deliberately sets no filesystem sandboxing, since `ProtectSystem` and friends break exactly the two things this app does.

**Upgrades are a symlink move.** The new release is unpacked beside the current one and `/opt/borg-ui/current` is repointed. `/var/lib/borg-ui` and `/etc/borg-ui/borg-ui.env` are never touched, so an operator's edits survive.

## Verification

Verified locally:
- All three shell scripts parse (`bash -n`) and the 6 new guard tests pass, along with the existing `test_workflow_python_version_parity` guard that my `release.yml` change had to keep satisfying.
- The Borg selection logic against the real manifest at glibc 2.31 / 2.36 / 2.41 / 2.43, confirming it picks the highest floor the host clears and that Borg 2 only resolves at 2.43.
- The tarball builder end to end with the frontend and agent-wheel steps stubbed (this machine has Node 20, the build needs 22): correct layout, correct `versions.env`, checksum written.

**Not verified, and it cannot be yet:** an actual install on a real Debian or Ubuntu host. `install.sh` downloads a release asset, and no published release carries one until this workflow change ships. The first release after this merges is the point to do a manual pass on a Debian 13 VM and a Proxmox LXC before pointing anyone at the one-liner.


karanhudia:
Addressed the review. Five findings were real and are fixed in 512cae37; one I am declining, with reasoning.

**Fixed**

- **Partial extraction poisoning the release directory** (`install.sh:154`). Real bug. The tarball now unpacks under the temporary directory and is renamed into place, so `RELEASE_DIR` only ever exists complete.
- **Upgrade ignoring the persisted `DATA_DIR` and `PORT`** (`install.sh:404`). Real bug, and the worst of the six: re-running over a `--data-dir /srv/borg-ui` install provisioned `/var/lib/borg-ui`, never fixed the real data directory's permissions, and printed the wrong URL. `load_persisted_settings` now reads both back from the env file before anything uses them, and says so if a flag contradicts what is already installed.
- **Release tree writable by the service user** (`install.sh:441`). Correct. The tree is now root-owned and read-only to that user, and the virtualenv with it. Nothing in the app writes inside the release tree, since every path it writes to derives from `DATA_DIR`.
- **Mutating the venv under a live process** (`install.sh:396`). The hazard is real. I took a smaller fix than the suggested per-release virtualenv: stop the service before pip runs. That removes the concurrent-mutation window entirely, and an upgrade already takes a restart, so the cost is a few seconds of downtime rather than a doubled install footprint and a rollback scheme nothing yet drives.
- **Uninstall leaving dangling links.** Fixed, and it only removes `/usr/local/bin/{borg,borg2,rclone}` where they still point into `/opt/borg-ui`, so a link the operator put there survives.

Also took the zizmor `template-injection` hint on the changelog step: its expansions now go through `env`.

**Declined: replacing `curl | sudo bash` from `main` with a signed installer asset**

The suggested mitigation is circular. Whatever verifies the bootstrap script has to be fetched over the same channel from the same repository, so a repository compromise defeats the checksum as readily as the script. `curl | sudo bash` over HTTPS from the project's own repository is the accepted pattern here (rustup, Docker, k3s, the Proxmox community scripts), and the trust boundary is GitHub either way.

One part of that finding was genuinely actionable and is fixed: the release notes pinned `--version` while still fetching `install.sh` from `main`, so a released version could be installed by a newer installer. The notes now name the installer from the tag, which is immutable, and the docs show that pinned form alongside the download-and-review example. The release archive itself was already SHA-256 verified before anything is unpacked.

karanhudia:
All four fixed in 932af409.

**The venv finding was right, and I was wrong to decline it the first time.**

Two findings converged on the same thing, and taken together they made the constraint obvious: with one shared virtualenv, an upgrade can avoid rewriting an environment a live process is importing from, *or* avoid holding the service down across the whole install, but not both. The previous commit picked the second, which is what the transactional finding then landed on.

One virtualenv per release resolves both. Download, unpack and environment build now happen beside the running service, which keeps serving from its own tree, so anything that fails during preparation exits with the old release still up. Only then comes the switch: stop, move the symlink, start. If the new release will not come up, the installer points `current` back at the previous release and starts that before reporting the failure. The previous release is retained for exactly that; older ones are pruned.

**Also fixed**

- **Cross-filesystem `mv`.** Correct, my fix was incomplete. Staging now happens under `/opt/borg-ui/releases`, so the move into place is a rename rather than a copy an interruption could truncate.
- **`SERVICE_USER` lost on upgrade.** Real, and the same bug class as the `DATA_DIR`/`PORT` one from the last round. A re-run without `--service-user` reset it to root, rewrote the unit and handed the data directory to root, silently undoing a deliberate restricted install. It is now read back from the installed unit.
- **Env file parsed unlike systemd.** Real. `DATA_DIR="/srv/borg-ui"` was taken with the quotes as part of the path, so the installer provisioned and chowned a directory the service does not use. Now last-assignment-wins with one layer of matching quotes stripped, matching how `EnvironmentFile=` reads it. Escape syntax is refused outright rather than guessed at, since a misparsed `DATA_DIR` chowns the wrong tree.

Two rounds of review have now found real bugs in that parsing, so it is no longer trusted by inspection: four tests extract `load_persisted_settings` and run it against quoted values, later-wins, the unit's `User=`, a contradicting flag, and unsupported escapes. Ten guard tests in total, all passing.

karanhudia:
@coderabbitai review

karanhudia:
@coderabbitai full review

karanhudia:
Addressed the last outstanding finding in ad4c747b. I was wrong to leave it.

My argument was that verifying a bootstrap script is circular when the checksum arrives over the same channel. That much is true, but it was an argument for doing nothing about a weakness that was real and separate: `raw.githubusercontent.com/.../main/scripts/install.sh` resolves to whatever `main` holds at the moment the command runs, and `main` moves after every merge. So the script a user piped into `sudo bash` was not the one that shipped with the release they were installing, quite apart from any compromise.

The installer is now published as an asset of each release with its SHA-256 beside it, and every documented command names one of those:

```bash
curl -fsSLO https://github.com/karanhudia/borg-ui/releases/latest/download/install.sh
curl -fsSL https://github.com/karanhudia/borg-ui/releases/latest/download/install.sh.sha256 | sha256sum -c -
sudo bash install.sh
```

What that does and does not buy, stated plainly: it does not defend against a compromise of this repository, and nothing fetched over a single channel can. What it fixes is the part that was broken. The artifact is fixed per release instead of following a branch, the checksum is published as a separate object from the script, and `sha256sum -c` exits non-zero before anything runs as root. Release notes carry the verifying form filled in for that release, and a pinned-to-a-release variant is documented for installing something other than the newest. A convenience one-liner remains, now pointing at the same asset and labelled as the unverified shortcut.

Verified the round trip locally: the builder emits `install.sh` and `install.sh.sha256`, and `sha256sum -c install.sh.sha256` prints `install.sh: OK`, which is the exact command the docs hand to users.

Two guard tests keep it from regressing: no documented install command may name `raw.githubusercontent.com`, and the release must publish both the installer and its checksum. Twelve guard tests now, all passing.

Also rebased onto the `main` merge you pushed.

karanhudia:
All four fixed in 5b804cc1. The critical one was mine and it was as bad as it looks.

**`main()` called a function that no longer existed.** Folding `stop_service_for_upgrade` into `activate_release` deleted the definition and left the call. Under `set -e` every install died at that line with `command not found`, before it created a single directory. The installer was completely non-functional, and twelve guard tests passed over it, because not one of them runs `main()`.

That gap is now closed by bash rather than by inspection: the script is sourced with its invocation stripped and every step `main()` names is resolved. I verified the guard is not vacuous by reintroducing the deleted call, which fails it, and removing it again, which passes.

**The other three**

- **Completion marker for the virtualenv.** Correct. pip installs in its own order, so an interrupted run could leave `bin/gunicorn` present while later dependencies were missing, and the next run would call that finished and activate it. A marker is now written only after the requirements install returns, and that is what is tested.
- **`--no-start` caused an outage.** Correct, and an ugly one: it stopped the service at the switch and then returned without starting it, so an upgrade with that flag took the service down until somebody noticed. It now stops only when this run will start it again. The running service keeps serving the code it already loaded, and the new release takes effect on the next restart, which is what the flag should have meant all along.
- **`systemctl restart` succeeding is not the app working.** Correct. `Type=exec` returns as soon as gunicorn is exec'd, before it binds the port, and says nothing about whether it stayed up, so a release dying on a bad migration or a taken port counted as a good start and never rolled back. The installer now waits for the app to answer on its port before declaring success, and holds the rollback to the same standard rather than assuming the old release came back up.

Thirteen guard tests, all passing. Worth saying plainly: three rounds of review have now found real defects in this script, including one that broke it outright, and the end-to-end install still cannot be exercised until a release publishes the tarball. A manual pass on a Debian 13 VM and a Proxmox LXC before this is advertised is not optional.

karanhudia:
Fixed in 7084a51e. Correct, and it defeated the rollback in the likelier of the two failure modes.

`wait_until_serving` fails two ways. If the unit goes inactive, `systemctl start` brings the previous release up and the rollback worked. But a release that stays *active* while never answering on its port is the other case, and the more probable one: a hung migration, a worker crash-looping inside a still-running unit. `start` is a no-op against an active unit, so `current` went back to the previous release while gunicorn carried on running the new one, and the installer reported a rollback that had not happened. `restart` covers both.

A guard keeps it that way, because the difference is one word and reads like a simplification: the installer may not bring the service up with a bare `systemctl start`. Verified it fails when the old form is put back.

Fourteen guard tests, all passing.

karanhudia:
@coderabbitai full review

karanhudia:
Both fixed in 0bd7f434.

**Service PATH.** Correct, and the more consequential of the two. The unit set no `PATH`, so the service inherited systemd's default and resolved `borg`, `borg2` and `rclone` from `/usr/local/bin` and `/usr/bin`. Since the installer deliberately leaves a pre-existing `/usr/local/bin/borg` alone, that file silently decided which Borg the service ran, and it need not be the version this release was built against. Running a different Borg than the server expects is exactly the drift the pinned-binary design exists to prevent, so this undercut it at the last step.

The unit now puts `@PREFIX@/bin` first. That also turns the pre-existing-binary conflict into a non-event rather than a warning to live with: the service uses its own Borg, the operator's shell keeps resolving to theirs. The warning text said the opposite and is corrected, and the docs now name `/opt/borg-ui/bin` in the install table.

**Header command.** Correct, and a plain miss. The installer's own header still advertised piping a `main` download into a root shell, after I had changed README and the docs. The guard meant to prevent that only read `README.md` and `docs/installation.md`, so it missed the file it was guarding. It now reads the installer too.

Both new guards were checked against the defects they describe rather than assumed: `PATH` without the prefix fails, the branch download in the header fails. Fifteen guard tests, all passing.

karanhudia:
@coderabbitai full review

karanhudia:
Both fixed in b63fe31e. The first one shows an argument I made in this PR was wrong.

**Checksum did not gate execution.** The documented blocks put `sha256sum -c -` and `sudo bash install.sh` on separate lines, so pasting a block ran the installer as root whether or not the checksum matched. That makes publishing the checksum decorative. I wrote in an earlier thread that the verification "exits non-zero before anything runs as root", which is only true if the commands are chained, and they were not. They are now chained with `&&` in the README, the docs, the installer's own header and the generated release notes.

**Stale forwarders.** Also correct, and it is a consequence of the PATH change in the previous commit: putting `@PREFIX@/bin` first on the service PATH turned every fallback into a way to leave a lie on disk.

- Falling back to the distribution `borgbackup` wrote nothing to `@PREFIX@/bin`, so a forwarder from an earlier install kept shadowing the package that had just been installed.
- `--skip-borg2` and a failed Borg 2 build left their forwarder in place, so Borg 2 looked available on a host the installer had just reported it unavailable on. That one is worse than a missing binary, because the operator was told the opposite.

Every branch that ends without a Borg now clears its forwarder, and the Borg 1 fallback writes one pointing at `/usr/bin/borg` by absolute path, since resolving `borg` at that point could find our own symlink and build a loop.

Both guarded. Seventeen guard tests, and each new guard was run against the defect it describes rather than assumed correct.

karanhudia:
@coderabbitai full review

karanhudia:
Both fixed in 2924b480.

**Tag names reaching shell source.** Correct. The changelog step expanded `steps.previous_tag.outputs.tag` and `github.ref_name` directly into its `run:` body. Worth spelling out why the previous tag is the sharper end of it: the release tag at least passes `check-release-version.sh`, while the previous one is picked off `git tag --sort=-v:refname` and never validated at all.

Every ref-derived value in the workflow now goes through `env:` and is referenced as a quoted shell variable, in the two earlier steps as well as the changelog, so the run bodies contain no expansions at all. A guard fails on any `${{ }}` left inside one.

**Options without values.** Correct. A trailing `--version`, `--port`, `--data-dir` or `--service-user` read `$2` under `set -u` and died with `unbound variable`, which tells an operator nothing about what they got wrong. Each now checks for its value and exits with a usage error naming the option.

That one comes with better tests than most of this file has had. Argument handling runs before anything privileged, so these paths can actually be executed rather than pattern-matched: the tests run the installer with each option empty, with `--help`, and with an unknown option, and assert the exit codes and the messages. That is real coverage of real behaviour, which is more than I can say for the parts of this script that need a Debian host.

Twenty guard tests, and both new guards were run against the defect they describe.

karanhudia:
@coderabbitai review

karanhudia:
@coderabbitai full review

karanhudia:
Both fixed in 2b1a5095.

**Service-user change orphaned the data.** Correct, and the sharpest of the recent ones. `install -d` owns the directory entries it creates, not what is already inside them, so passing `--service-user` to an existing install rewrote the unit with the new `User=` while the database, the secret key and every deployed SSH and Borg key kept the old owner. The new service could not open any of them. Rollback does not cover it, as you say: it moves the `current` symlink while the unit still runs as the new user. The installer now detects the change and re-owns the data directory recursively, and warns that it may take a while since backups live under it by default.

**Unverified one-liner.** Fair. The label was in prose above the block, which is exactly what a reader skimming for a command skips. It now has its own heading naming it unverified and a comment inside the block itself, so the label travels with the command that gets copied.

Twenty-two guard tests, both new ones checked against the defect they describe.

---

A note on the shape of this review, since eight rounds is worth stepping back from. Of the twenty-three findings so far, roughly ten are the same class: state left inconsistent by a **second** run, an **interrupted** run, or a run with **different flags**. Partial extraction, the shared virtualenv, persisted `DATA_DIR` and `PORT`, the service user, stale forwarders, the venv completion marker, and now data ownership. The installer was written as a linear first-install script, but it is really a convergence tool acting on a host that already has state, and each fix has been local to one symptom.

Every one of those is reachable by running the installer twice. None need exotic conditions. The reason none of them are caught here is that nothing in CI executes this script: the twenty-two guards are structural, plus argument parsing, and cannot catch a state bug by construction.

The missing piece is an integration job that runs the installer in a systemd-capable Debian container across that matrix: fresh install, plain re-run, re-run with a different `--data-dir`, `--service-user` change, `--skip-borg2` after Borg 2 was installed, `--no-start` against a running service, and a killed run resumed. That needs one product change to be possible, a way to install from a locally built tarball rather than a published release, which is worth having anyway for air-gapped installs. I have raised it with Karan rather than adding it to this PR unasked.

karanhudia:
The installer now runs, and the review pattern has a root cause.

Across eight rounds, roughly ten of the twenty-three findings were the same class: state left inconsistent by a second run, an interrupted run, or a run with different flags. Every one of those is reachable by running the installer twice, and nothing here had ever run it once. The guard tests are structural and cannot catch that class by construction.

`scripts/test-native-install.sh` now runs the installer against a throwaway Debian container with systemd as PID 1, across exactly those transitions:

```
==> Case: fresh install                                          ok x5
==> Case: plain re-run is idempotent                             ok x4
==> Case: a re-run must not override the installed data dir/port ok x4
==> Case: --no-start must not take a running service down        ok x2
==> Case: --skip-borg2 removes a borg2 it can no longer provide  ok x2
==> Case: a half-built virtualenv is rebuilt, not trusted        ok x2
==> Case: changing the service user re-owns the existing data    ok x5
==> 24 passed, 0 failed
```

It needed one product change to be possible: `--tarball` installs from a locally built artifact rather than a published release, which is also how an air-gapped host installs. Without it the installer could only fetch a release asset, which is why this PR had gone eight rounds with zero end-to-end verification.

**The first run found four defects that no amount of reading had.**

- `tar -tzf "$TARBALL" | head -n 1` killed tar with SIGPIPE once `head` had its line, and under `set -o pipefail` that failed the whole installer before it created a single directory. GNU tar does this; the BSD tar on a Mac does not, so it appeared only where the installer actually runs.
- Building the tarball on macOS wrote an AppleDouble `._name` sidecar for every file with an extended attribute. They are binary and they match `*.py`, so Alembic loaded `._<revision>_baseline_schema.py` as a migration and died on "source code string cannot contain null bytes". The build now inspects the tarball it just wrote, because a successful `tar` is not the same as an installable artifact.
- The image sets `OPENSSL_armcap=0` because cryptography's OpenSSL path dies with SIGILL on some ARM64 kernels. The native path did not. On ARM64 the app installed perfectly and then crash-looped on import before writing a row. Measured both ways in the container: exit 132 without it, clean with it. A parity test now fails whenever the image sets something the native path does not, which is the check that would have caught this without a container at all.
- On a fresh install, `readlink -f` resolved the not-yet-existing `current` symlink to itself, so a failed first start "rolled back" to the symlink it had just pointed at the failing release, and reported success at doing so.

The harness also refuses to run when a packaged source file is newer than the tarball. The installer takes the unit and the app from the tarball, so editing a source and re-running silently tests the previous build. That happened once here and made a correct fix look broken, which is worse than not testing.

Six bugs in the harness itself were found the same way, by running it.

karanhudia:
All six fixed in 5c60cb54, verified against a rebuilt tarball: 28 matrix assertions and 25 unit guards pass.

**Removing the active release during local staging.** Correct, and it undid this PR's own transactional design for the path I had just added. A local tarball is normally a rebuild of the version already installed, and the release directory was keyed by version alone, so staging deleted the tree the running service was using before the new virtualenv existed. A failure during preparation would have left `current` dangling with no previous release to roll back to. The directory is now keyed by version plus a digest of the tarball, so `current` stays intact through preparation, and rebuilding identical content reuses the directory rather than replacing it.

**Unknown `--case` reported success.** The sharpest of the minor ones. It matched nothing, ran nothing, printed `0 passed, 0 failed` and exited 0. Worse than you flagged: the harness's own header documented `--case upgrade`, which is not a case, so the documented example would have reported success having tested nothing. Unknown values now exit 2 and list the real cases; the header names them.

**Options without values** now exit 2 with the offending option named, matching `install.sh`.

**Freshness guard** now watches `build-native-tarball.sh`, `docker/runtime-base.env` and `Dockerfile.runtime-base`. All three change what the tarball contains, so a stale archive could pass the guard and be installed. This one matters more than it looks: testing a stale artifact already made a correct fix look broken once in this PR.

**Parity guard matched substrings.** Right, and the example is mine: the `OPENSSL_armcap` comment I added to the unit would have satisfied the guard on its own. It now parses actual assignments -- `Environment=` in the unit, `export` in the launcher, the env-file heredoc in the installer -- and I confirmed it fails when the assignment is replaced by a mention.

**Globbed `--tarball` in the docs** now resolves one explicit path first.

Worth noting what this round consisted of: five of the six were defects in the test harness and its guards rather than in the installer. That is the right direction for a review to be pointing after the installer itself started being executed, but it is also a reminder that a harness is code too, and mine has now had eight of its own bugs.

PR #973, opened 2026-09-08
fix(db): add the agent upgrade tracking columns as an Alembic revision

karanhudia:
## Problem

Any request that reads `agent_machines` fails on a live database:

```
sqlalchemy.exc.OperationalError: (sqlite3.OperationalError) no such column: agent_machines.desired_agent_version
```

The centralized agent upgrades work (#951) added six columns to `AgentMachine` in the model, but the only schema change it shipped was a legacy ladder file, `129_add_agent_upgrade_tracking.py`. That ladder is frozen and no longer runs at startup. It is replayed only on a throwaway copy of a pre-Alembic database, whose rows are then copied onto the Alembic baseline, and the baseline predates these columns. So no Alembic-managed database has ever had them.

Missing columns: `desired_agent_version`, `desired_borg_version`, `upgrade_state`, `upgrade_requested_at`, `upgrade_target_version`, `upgrade_error`.

The unit tests did not catch it because they build the schema from the model metadata rather than from the revision chain.

## Fix

A new revision `f7a2b8c4d1e5` on the current head `e5f6a7b8c9d0`, adding the six nullable columns and dropping them again on downgrade.

Tests cover the upgrade, the downgrade, and that the revision chains on the previous head and leaves a single head.

## CodeRabbit

One finding (major): `downgrade()` dropped columns that `upgrade()` might not have created, because the first version of this revision skipped columns that already existed.

Fixed by removing that guard rather than by tracking created columns. It was defending against a state `db_upgrade` cannot produce, since the copy target is always a freshly built baseline, and it was the only source of the asymmetry.

### DIFF
--- a/app/database/alembic/versions/f7a2b8c4d1e5_add_agent_upgrade_tracking.py
+++ b/app/database/alembic/versions/f7a2b8c4d1e5_add_agent_upgrade_tracking.py
@@ -1,0 +1,48 @@
+"""add agent upgrade tracking columns
+
+The columns landed in the model with the centralized agent upgrades work but
+only ever got a legacy ladder file (129), which is frozen and no longer runs at
+startup. Every Alembic-managed database is therefore missing them.
+
+Revision ID: f7a2b8c4d1e5
+Revises: e5f6a7b8c9d0
+Create Date: 2026-09-08
+"""
+
+from alembic import op
+import sqlalchemy as sa
+
+
+revision = "f7a2b8c4d1e5"
+down_revision = "e5f6a7b8c9d0"
+branch_labels = None
+depends_on = None
+
+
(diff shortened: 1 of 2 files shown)


karanhudia:
CodeRabbit flagged that `downgrade()` dropped columns `upgrade()` may not have created.

Rather than tracking created columns, I removed the skip-if-present guard from `upgrade()`. It was protecting against a case that cannot happen: `db_upgrade` catches a pre-Alembic database up on a throwaway copy and then copies its rows onto a freshly built baseline, which has none of these columns. The revision is now a plain symmetric add/drop, same as every other one in the tree.

Commit 718b07d4. 7 tests pass.

karanhudia:
@coderabbitai review

Issue #936, opened 2026-09-06
Agent: `repository.storage_usage` job that measures Borg 2 repositories on store URLs (replaces `du` on a URL)


ioanalytica:
## Summary

For Borg 2 repositories executed by a managed agent, the server's size refresh ends in the agent job `repository.disk_usage`, which runs `du -sb -- <repository_path>`. For every store URL (`sftp://`, `rest://`, `rclone:`, `s3:`) that is `du` on a string that is not a path; the job fails and `total_size` stays NULL. This issue proposes a new agent job kind `repository.storage_usage` that applies the same source order as the server side (#934) on the agent, reports a JSON result with the quantity and its source, and is advertised as a capability so a mixed fleet keeps working.

## Current behaviour

- Agent (`agent/borg_ui_agent/repository_ops.py`): `repository.disk_usage` builds `["du", "-sb", "--", repository_path]`; the result goes back through `_execute_short_repository_operation` as `{return_code, stdout, stderr, data}`.
- Server (`app/api/repositories.py::_update_agent_repository_stats`): Borg 1 takes `cache.stats.unique_csize` from `repository.rinfo`; Borg 2 falls back to `repository.disk_usage`, parses `<bytes>\t<path>` from stdout, and leaves `total_size` unchanged on any failure.
- Every enrolled agent 0.1.3 lists `repository.disk_usage` in `capabilities`; the server checks `job_kind in agent.capabilities` before queueing (`repository_executor.py`, `validate_agent_repository_operation`).

## What the agent can measure (verified on Borg 2.0.0b23 and 2.0.0b24)

1. **Chunk-index sum** (exact, Borg's own notion of repository size): `Repository(Location(url), exclusive=False, lock=False).list()` yields `(chunk_id, storage_size)` from the store's `index/` fragments; the sum equals `compact --stats`'s "Repository size" byte for byte, needs no passphrase, loads no pack, and reads while a backup holds the exclusive lock. It requires an importable `borg`. In the container agent image the configured Borg 2 binary resolves to a venv (`readlink -f $BORG2_BINARY` -> `<venv>/bin/borg`), and `<venv>/bin/python -c "import borg, borgstore"` succeeds. A PyInstaller Borg 2 (bare-metal agents installed by `install.sh`) has no interpreter next to it.
[...]

Issue #935, opened 2026-09-06
Repository status from repository evidence: backup from archives, prune from detected removals, one model for card, strip and info dialog


ioanalytica:
## Summary

The repository card shows two answers to "when did this repository last get a backup": the key stat `Last backup` (from `repositories.last_backup`, derived from the archives table) and the status strip's `Backup` cell (from the newest `backup` operation or `BackupJob`). For a repository whose backups are created outside Borg UI (a cron job, the CLI, another Borg UI instance) the card says "today" and the strip says "x days ago, overdue" on the same screen. The strip's `Prune` and `Compact` cells behave the same way. This issue proposes one server-side status model per repository that uses the best available evidence per cell (the repository where Borg provides it, job rows otherwise) and is rendered by one component for Borg 1 and Borg 2.

## Current behaviour

`GET /repositories/{id}/status-strip` (`app/api/archive_index.py`, #902): per cell the newest terminal row of `operations` (kinds `backup` / `check` / `prune` / `compact`, categories `index` / `mirror`) unioned with the legacy job tables (`legacy_status.py`), newer wins. No cell reads the repository. Overdue is `now - completed_at > OVERDUE_THRESHOLD_DAYS[cell]` with fixed thresholds (backup 2, prune 14, compact 30, check 30, index 2, mirror 1 days); with no row at all `overdue` is true.

Observed on a repository whose backup, prune and compact run from a cron job: `last_backup` = today (archives table), strip `Backup` / `Prune` / `Compact` = the last run that went through Borg UI, weeks earlier, flagged overdue. The archives table had registered the new archive and `archive_sync` had reported the pruned one as removed the same day.

Related symptoms:

- `repositories.last_backup` has three writers (`write_repository_archive_columns`, `_finish_linked_backup_job`, `_update_agent_repository_stats`, plus `backup_service._update_repository_stats`), noted in the phase 1 status as "to be resolved".
- The info dialog renders two components (`RepositoryStatsV1` / `RepositoryStatsV2`) with different field sets from two JSON shapes.
- Spec 10.2 describes the strip as job-driven; spec 6.4 already makes `Archive.backup_operation_id` nullable, i.e. archives without a Borg UI operation are a modelled case.

## What the repository can tell us (Borg 1.4.5, 2.0.0b23 and 2.0.0b24, measured)

| Fact | Source | Both versions |
|---|---|---|
| last backup | `max(archives.start)` (archive list, already synced) | yes |
[...]

Issue #934, opened 2026-09-06
Borg 2 repository size and `last_modified`: read-only source order in `stats`, persisted `borg_last_modified`


ioanalytica:
## Summary

For Borg 2 repositories on store URLs (`sftp://`, `rest://`, `rclone:`, `s3:`) `repositories.total_size` stays NULL, because every size path ends in `du` on a URL. Borg 2 reports no size through any command (borgbackup/borg#10329 asks for one), but the number is read-only available in the repository's chunk index and reproduces `compact --stats` exactly. Separately, both Borg versions report `repository.last_modified`, and Borg UI neither persists it nor returns it consistently. This issue proposes one `stats` source order with an explicit source label, and a `borg_last_modified` column.

## Current behaviour

Size, by path:

| Path | Code | Borg 1 | Borg 2 |
|---|---|---|---|
| server `stats` | `BorgRouter.calculate_total_size_bytes` | `info --json` `cache.stats.unique_csize` | `RepositoryV2Service.calculate_total_size_bytes`: `du` over ssh when `repository.host` is set, else local `du` on `repository.path`; fails for every store URL |
| agent `stats` | `_update_agent_repository_stats` | `repository.rinfo` `cache.stats` | `repository.rinfo` has no size, fallback `repository.disk_usage` = `du -sb <url>`; fails for every store URL |

On a failure `total_size` is left unchanged, so a repository that never had a size shows `N/A` forever.

`last_modified`:

| Consumer | Value today |
|---|---|
| `GET /repositories/{id}/info` (info dialog) | Borg's own `repository.last_modified` from the raw `info` / `repo-info` JSON, both executors |
| `GET /repositories/{id}` and `GET /repositories/{id}/stats` (`get_repository_stats`) | agent repositories: `repository.updated_at` (a Borg UI row timestamp); server repositories: `None` |
| database | not stored anywhere |

## Measured (Borg 2.0.0b23 and 2.0.0b24, identical results)

- The chunk index lives in the store's `index/` fragments and is maintained by every write. `Repository.list()` yields `(chunk_id, storage_size)`; the sum equals compact's "Repository size" byte for byte (301292 bytes vs "301 kB in 4 objects"), without loading a pack.
- It needs no key: works on an `aes256-ocb` repository with `BORG_PASSPHRASE` unset.
[...]

Issue #933, opened 2026-09-06
Enqueue `archive_sync` when a backup completes, so the archive index does not wait for the next reconcile run


ioanalytica:
## Summary

Since phase 1 of the operations spec, the archives table (and the columns derived from it: `archive_count`, `last_backup`, the status strip's Index cell, the heatmap, removal detection for pruned archives) is written only by `archive_sync`, which is enqueued by the reconcile scheduler every `stats_refresh_interval_minutes`. A backup that finishes does not enqueue anything. The spec plans a `backup` follow-up chain in phase 8 and lists the legacy stats refresh after a backup as "replaced by a one-line `enqueue` in phase 5" (section 15 table, `backup_service.py:2643` / `:2759`). Until then the index lags the repository by up to one reconcile interval, and the two backup completion paths keep writing `last_backup` themselves.

## What happens today

| Path | On completion | Archives table |
|---|---|---|
| Server backup (`app/services/backup_service.py`, `_update_repository_stats`) | runs its own `borg list`, writes `archive_count` / `last_backup` on the repository row | untouched |
| Agent backup (`app/api/agents.py`, `_finish_linked_backup_job`) | `repository.last_backup = completed_at` | untouched |
| Reconcile (`app/services/operations/reconcile.py`) | `archive_sync -> history_merge -> history_index -> stats` per repository | updated |

Observed on an installation with `stats_refresh_interval_minutes = 720`: a repository shows `last_backup` 09:46 (from the job) while `max(archives.start)` is 07:46 from the previous day's reconcile run, the status strip says "Index about 14 hours ago", and the removal of the archive the same plan run pruned is not detected until the next reconcile. The timer also restarts with the process, so a restart shortly before the next run pushes it out by another interval.

`FOLLOWUPS["backup"]` already exists (`("archive_sync", "history_index", "stats")`, `app/services/operations/followups.py`) and is unused.

## Proposal

At backup completion (`completed` or `completed_with_warnings`), in both paths, enqueue the follow-up chain instead of the ad-hoc stats refresh:

```python
from app.services.operations.enqueue import enqueue_chain
from app.services.operations.followups import chain_for, history_enabled
from app.services.operations.reconcile import has_active_index_work

if not has_active_index_work(db, repository.id):
    enqueue_chain(
        db,
        chain_for("backup", available=registered_kinds(), history=history_enabled(db)),
        repository_id=repository.id,
        trigger="followup",
        backup_plan_run_id=...,   # when known
    )
```

[...]

PR #974, opened 2026-09-08
feat(borg2): run compact with --stats and persist the repository statistics

### LINKED ISSUE #931: Borg 2: run every compact with `--stats -v`, parse and persist the repository statistics it reports
## Summary

Borg 2 reports repository-wide statistics in exactly one place: `borg compact --stats`, on INFO level. Borg UI runs compact for Borg 2 repositories in three code paths and none of them passes `--stats`, so the only numbers Borg 2 offers for "how big is this repository, how much source data, how well does it deduplicate" are thrown away on every run. Adding the flag and parsing the lines gives Borg 2 repositories the fields the info dialog shows for Borg 1 from `cache.stats`, with a "as of last compact" timestamp.

## What Borg 2 prints (verified on 2.0.0b23 and 2.0.0b24)

`borg compact --stats -v` logs (INFO, logger `borg.archiver.compact_cmd`):

```
Overall statistics, considering all 2 archives in this repository:
Source data size was 1 MB in 6 files.
Deduplicated size is 500 kB.
Deduplication factor is 0.50.
Repository size is 502 kB in 6 objects.
Compression factor is 1.00.
Compaction saved 0 B.
```

With `--log-json` each line arrives as `{"type": "log_message", "levelname": "INFO", "name": "borg.archiver.compact_cmd", "message": "..."}`. The values are rounded to three significant digits (`format_file_size(..., precision=0)`); Borg has no JSON output for them (borgbackup/borg#10329 asks for a read-only `repo-info --stats`). `--stats` is not honoured with `--dry-run`, and the maintainer notes in borgbackup/borg#8796 that `--stats` makes compact somewhat slower.

`info --help` in b23 and b24: "for the repository-wide deduplicated size, use borg compact --stats."

## Current code

| Path | Command today | Verbosity |
|---|---|---|
| `app/services/v2/compact_service.py` (server, Borg 2) | `borg2 -r <repo> compact --progress --log-json` | none: INFO lines are not emitted |
| `app/core/borg2.py` `Borg2Interface.compact` | `borg2 -r <repo> compact` | none |
| `agent/borg_ui_agent/repository_ops.py` `repository.compact`, Borg 2 branch | `borg2 -r <repo> compact --progress --verbose --log-json` | INFO emitted, but no `--stats` |

The server compact service already parses `--log-json` stderr for `progress_percent` (`compact_service.py` around line 265); the agent streams the same lines into the job log. Borg 1 `compact` has no `--stats` and is not affected.

## Proposal

1. Add `--stats` (and `--info`, where missing) to the three Borg 2 compact commands.
2. Parse the five `log_message` lines from logger `borg.archiver.compact_cmd` into

   ```json
   {
     "archive_count": 2,
     "source_size": 1000000, "source_files": 6,
     "deduplicated_size": 500000, "deduplication_factor": 0.50,
     "repository_size": 502000, "object_count": 6,
     "compression_factor": 1.00,
     "compaction_saved": 0,
     "unit_precision": "3_significant_digits"
   }
   ```

   Sizes converted from Borg's decimal units (kB = 1000) to bytes as printed, without pretending more precision than the text carries.
3. Persist the object on the compact job (`compact_jobs.stats`, one nullable JSON column) for the legacy path, and on `operations.result` once phase 5 of the operations spec moves compact there; for agent compacts, parse in `_finish_linked_repository_operation_job` from the collected job logs, the same way `archive_names_from_prune_output` reads prune output today.
4. Refresh `repositories.total_size` from `repository_size` when present (labelled as coming from the last compact, so the UI can say "as of <compact time>"), and expose `source_size`, `deduplicated_size`, `compression_factor` to the repository-info dialog for Borg 2, which today shows nothing in these slots.

## Why this matters

- Two live installations with Borg 2 `rest://` repositories run a compact after every backup plan run and still show `Total size: N/A`.
- Borg 1's info dialog shows used-on-disk, unique data, chunk counts from `cache.stats`; Borg 2 has no client-side equivalent except this output.
- Without the flag the numbers are lost at the only moment Borg computes them.

## Out of scope

A read-only, on-demand size (chunk-index sum, `repo-info --stats` from borgbackup/borg#10329) is a separate topic; this issue is only about not discarding what compact already reports.

ioanalytica:
## Description

Borg 2 reports repository-wide statistics in exactly one place: `borg compact --stats`, on INFO level (`info --help`: "for the repository-wide deduplicated size, use borg compact --stats"). The compact service never passed the flag, so the only size Borg 2 offers was discarded on every run and Borg 2 repositories kept showing `Total size: N/A`. Server half of #931; the agent half (agent 0.1.4 runs `compact --stats` under `BORG_UNITS=raw`) landed with #968. Built on the `operations` substrate of #947.

- **Parser** `app/services/borg2_compact_stats.py`: reads the statistics lines (verbatim 2.0.0b23 / b24 output, plain or wrapped as `--log-json` `log_message` entries of any logger) into `{archive_count, source_size, source_files, deduplicated_size, deduplication_factor, repository_size, object_count, compression_factor, compaction_saved, size_precision}`. Every compact path sets `BORG_UNITS=raw`, so sizes are exact byte counts; `size_precision` records when they are not. Agent-supplied bytes cannot raise out of it.
- **Command**: the server compact service runs `compact --stats --info` and warns when a successful compact printed no statistics. Borg's WARNING and ERROR frames stay at info in the server log; the `--info` chatter goes to debug.
- **Persistence**: the statistics live in the operation's `result["stats"]` (spec 6.1), reached through a `stats` property on `MaintenanceJobFacade`. No migration. The runner keeps the `stats` key the row holds when it writes its executor's result, read back from the table; nothing else on a row survives a requeue. `apply_compact_completion` stores the dict on success and warnings and refreshes `repositories.total_size` from `repository_size` with `total_size_source = compact_stats` (0 included: an emptied repository is a measurement). A rounded figure never replaces a measured size.
- **Agent path**: completion parses the collected job log. A line that arrives within ten minutes after the job's server-stamped completion (the agent's outbox drains after the completion report; seen live 150 ms apart) is appended to the operation's transcript, a failed run's error lines included; for a successful compact a closing statistics line triggers one parse of the transcript's tail. The size is refreshed only when no `stats` follow-up has measured the repository since (both `run_stats` branches now report `size_refreshed`) and no later compact reported one; the lookback is bounded below by the job's server-stamped claim, so the agent's clock cannot widen it.
- **API**: `serialize_job_status` includes `stats` for compact jobs; `/api/operations` carries it in `result` as before.
- `docs/architecture/job-system.md` describes the behaviour and its limits.

Not in this PR: the read-only size source order (#934) and the UI for the new fields (#935). A prune's late `--list` lines reach the transcript but are not fed back into the pruned-archive marking; a size refresh a user triggers by hand writes no operation row and is not seen by the guard.

## Related Issue

Fixes #931

## Type of Change

- [x] ✨ New feature (non-breaking change which adds functionality)
- [x] 📖 Documentation update
- [ ] 🎨 UI/UX improvement
- [ ] ♻️ Refactoring

## Testing

New and changed tests:

- `test_borg2_compact_stats.py`: verbatim b23/b24 output from plain lines and `--log-json` entries; junk, other loggers and nesting past the recursion limit; `None` without a repository-size line; decimal, raw and IEC units; the closing-line classifier.
- `test_operations_job_facade.py`: `stats` round-trips through `result["stats"]` next to other keys.
- `test_operations_runner.py`: `_merged_result` keeps `stats` and nothing else; a merge with a key another session filed while the executor ran.
- `test_operations_index_executors.py`: both `run_stats` branches report `size_refreshed`.
- `test_maintenance_state.py`: stats persisted and `total_size` refreshed on success and warnings; untouched without stats, on failure, for a rounded or unlabelled figure over a measured size, and with `refresh_size=False`.
- `test_v2_services.py`: `execute_compact` on an `Operation` row spawns `compact --stats --info` under `BORG_UNITS=raw`, persists `result["stats"]` (exact) and `total_size`, and keeps both across a `commit_with_retry` rollback.
- `test_api_agents.py`: completion parses the job log, guarded against a `stats` follow-up that measured while the job was unreported; late lines over the HTTP route and the session handler (the whole block after completion, a two-line frame, a failed compact's error lines, the 9/11-minute window boundary, a resend, an absorb that raises, a missing transcript file, a legacy `compact_jobs` row); the size guard against a `stats` follow-up that measured, one that could not, a later compact, and a slow agent clock.

Runs (scratch venv, Python 3.12, ruff 0.16.5):

```
pytest tests/unit/test_borg2_compact_stats.py tests/unit/test_maintenance_state.py \
       tests/unit/test_v2_services.py tests/unit/test_api_agents.py \
       tests/unit/test_operations_runner.py tests/unit/test_operations_index_executors.py \
       tests/unit/test_operations_job_facade.py tests/unit/test_operations_maintenance_executors.py \
       tests/unit/test_api_maintenance_jobs.py tests/unit/test_api_repositories.py
387 passed

pytest tests/unit   (full suite, untouched export of this branch on main)
3670 passed, 13 skipped

ruff check .            All checks passed!
ruff format --check     clean
```

Measured on Borg 2.0.0b23 and 2.0.0b24 (identical): `compact --stats -v` prints the statistics block; `compact --dry-run` answers "Ignoring --stats"; `create --stats` reports store traffic only. The parsed `repository_size` equals the chunk-index sum within Borg's rounding.

- [x] All existing tests pass

## Checklist

Not applicable.

## Additional Context

`--stats` makes compact somewhat slower (borgbackup/borg#8796), because Borg analyses the archives; the per-archive reference caches Borg keeps in the repository make unchanged archives cheap. A read-only `repo-info --stats` is requested upstream as borgbackup/borg#10329; until then this is the only Borg-native source. Part of the wave described in #917, #931, #933, #934, #935, #936.

### DIFF
--- a/agent/borg_ui_agent/compact_stats.py
+++ b/agent/borg_ui_agent/compact_stats.py
@@ -1,0 +1,243 @@
+"""Repository statistics from `borg compact --stats` (Borg 2).
+
+Borg 2 reports repository-wide statistics only here, on INFO level; there
+is no JSON form (borgbackup/borg#10329). The parser accepts the plain lines
+or the `--log-json` `log_message` entries that wrap them.
+
+Sizes are printed through Borg's `format_file_size`, which follows
+`BORG_UNITS`: `si` (default, `502 kB`, `2.39 TB`), `iec` (`490 KiB`) or
+`raw` (`502000 B`). The human forms are rounded to the digits Borg
+prints, so the parsed byte count is only as exact as those digits; every
+compact path of this application sets `BORG_UNITS=raw` and gets exact
+integers. The result records which it got in `size_precision`, from the
+repository size token, the one figure that is acted on: `exact` for a raw
+byte count, `rounded_to_printed_unit` otherwise. The deduplication and
+compression factors are printed with two decimals in either mode.
+
+A figure Borg cannot have printed (a size or count at or above 2**63, a
+factor that is not a finite number, a digit run `int` refuses) makes its
+line not a statistics line; the same parser reads agent-supplied
+transcripts, so nothing here raises on a line's content.
+
+The lines are matched by their whole wording, not by the logger that
+emitted them (`borg.archiver.compact_cmd` today); a `--log-json` entry of
+any logger is accepted.
+
+This is the agent's copy of `app/services/borg2_compact_stats.py`: the
+agent wheel ships only its own package, and the server keeps the parser
+for its own compacts and as the fallback for an agent that reported no
+statistics. Keep the two the same; a unit test compares them.
+(`is_stats_line` is unused here; it stays so the bodies are identical.)
+"""
+
+import json
+import re
+from decimal import ROUND_HALF_UP, Decimal
+from typing import Iterable, Optional
+
+PRECISION_EXACT = "exact"
+PRECISION_ROUNDED = "rounded_to_printed_unit"
+# A size or count at or above this did not come from Borg (its sizes are
+# 64-bit); a transcript line carrying one is not a statistics line.
+MAX_COUNT = 2**63
+# More digits than that cannot stay below MAX_COUNT, and `int` refuses a
+# long enough digit run outright (sys.int_info.str_digits_check_threshold).
+_MAX_DIGITS = 24
+# A Borg 2 compact prints its statistics as the last lines of the run; this
+# many lines from the end of the output hold all of them.
+TAIL_LINES = 64
+# The Borg 2 beta that added `compact --stats`; an older beta rejects the
+# whole command for the flag.
+COMPACT_STATS_SINCE_BETA = 15
+
+
+# Bounded components: `int` on an unbounded digit run can refuse it.
+_VERSION_TOKEN = (
+    r"(?P<major>\d{1,6})\.(?P<minor>\d{1,6})\.(?P<patch>\d{1,6})(?:b(?P<beta>\d{1,6}))?"
+)
+_VERSION = re.compile(r"\b" + _VERSION_TOKEN)
+_BORG_VERSION = re.compile(r"\bborg\S*\s+" + _VERSION_TOKEN)
+
+
+def parse_borg_version(output: str) -> Optional[str]:
+    """The version token in `borg --version` output ("borg2 2.0.0b24"):
+    the one after the program name where there is one (a wrapper may print
+    its own banner first), else the first version-shaped token, else None;
+    the caller then knows nothing about the binary rather than something
+    wrong."""
+    match = _BORG_VERSION.search(output) or _VERSION.search(output)
+    if not match:
+        return None
+    return match.group(0)[match.start("major") - match.start(0) :]
+
+
+def has_compact_stats(version: str) -> bool:
+    """Whether Borg `version` (a token from `parse_borg_version`) accepts
+    `compact --stats`: any 2.x that is not a beta before b15."""
+    match = _VERSION.search(version)
+    if not match or int(match.group("major")) != 2:
+        return False
+    beta = match.group("beta")
+    return beta is None or int(beta) >= COMPACT_STATS_SINCE_BETA
+
+
+_UNITS = {
+    "B": 1,
+    "kB": 10**3,
+    "MB": 10**6,
+    "GB": 10**9,
+    "TB": 10**12,
+    "PB": 10**15,
+    "EB": 10**18,
+    "KiB": 2**10,
+    "MiB": 2**20,
+    "GiB": 2**30,
+    "TiB": 2**40,
+    "PiB": 2**50,
+    "EiB": 2**60,
+}
+_SIZE = r"(?P<size>\d+(?:\.\d+)?) (?P<unit>[kMGTPE]?B|[KMGTPE]iB)"
+_PATTERNS = {
+    "archive_count": re.compile(
+        r"^Overall statistics, considering all (?P<n>\d+) archives in this repository:$"
+    ),
+    "source": re.compile(rf"^Source data size was {_SIZE} in (?P<n>\d+) files\.$"),
+    "deduplicated_size": re.compile(rf"^Deduplicated size is {_SIZE}\.$"),
+    "deduplication_factor": re.compile(
+        r"^Deduplication factor is (?P<f>\d+(?:\.\d+)?)\.$"
+    ),
+    "repository": re.compile(rf"^Repository size is {_SIZE} in (?P<n>\d+) objects\.$"),
+    "compression_factor": re.compile(r"^Compression factor is (?P<f>\d+(?:\.\d+)?)\.$"),
+    "compaction_saved": re.compile(rf"^Compaction saved {_SIZE}\.$"),
+}
+
+
+def _bytes(match: re.Match) -> Optional[int]:
+    """Integer bytes without a float in between: `int` keeps raw counts
+    exact above 2**53, Decimal scales the human forms as printed. None
+    for a figure Borg cannot have printed."""
+    size, unit = match.group("size"), match.group("unit")
+    if len(size) > _MAX_DIGITS:
+        return None
+    if unit == "B" and "." not in size:
+        value = int(size)
+    else:
+        scaled = Decimal(size) * _UNITS[unit]
+        value = int(scaled.to_integral_value(rounding=ROUND_HALF_UP))
+    return value if value < MAX_COUNT else None
+
+
+def _count(match: re.Match) -> Optional[int]:
+    text = match.group("n")
+    if len(text) > _MAX_DIGITS:
+        return None
+    value = int(text)
+    return value if value < MAX_COUNT else None
+
+
+def _factor(match: re.Match) -> Optional[float]:
+    text = match.group("f")
+    # The pattern admits digits and one point only, so the digit cap is
+    # also what keeps the float finite.
+    return float(text) if len(text) <= _MAX_DIGITS else None
+
+
+def _message(line: str) -> Optional[str]:
+    """The statistics line inside a `--log-json` entry, or the line itself."""
+    text = line.strip()
+    if not text:
+        return None
+    if text[0] != "{":
+        return text
+    try:
+        entry = json.loads(text)
+    except (ValueError, RecursionError):
+        # Agent-supplied bytes: nesting past the interpreter's limit, or an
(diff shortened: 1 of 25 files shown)


ioanalytica:
Integrative rollout test based on current main + this one has also passed on x86_64 and aarch64

karanhudia:
Parser and the server compact path look good. Two questions before I merge.

1. The agent already runs `compact --stats` itself since 0.1.4. Could it parse its own output and send the stats in the completion report, instead of the server recovering them from late log lines? That would drop the absorb window, the size guard and the runner result merge. Is there a reason that does not work, beyond needing another agent release?

2. `compact_stats` is pack file bytes, `borg2_index` is object size, `storage_used` is store file bytes. `total_size` now holds whichever wrote last, so the number changes meaning between a compact and the stats follow-up. Is that what the source order in #934 expects, or should compact only fill it when nothing else can measure?

ioanalytica:
Thanks, both questions deserve numbers rather than a quick yes, so I measured before answering.

2. What total_size holds. By definition you are right: compact_packs() sums the pack files from store_list("packs"), and the source says a pack's size "also counts bytes no index entry covers" (same in b23 and b24), while borg2_index sums object sizes from the chunk index. In practice the two are the same number at the moment compact writes it: on seven Borg 2 repositories the repository_size from compact --stats (BORG_UNITS=raw) and the chunk-index sum measured a few minutes later agree to the byte, object counts included, on all seven (24 MB to 2.5 GB). The reason is that compact_stats is only ever written right after a compaction, when the packs hold no garbage; a difference can only appear from packs kept below the compaction threshold, and only upwards. So the number does not change meaning between the compact and the stats follow-up in any case I can produce; it changes label.

That said, the source order in #934 is "measurement first", and FOLLOWUPS["compact"] is ("stats",), so every compact is followed by a measurement seconds later anyway. The clean rule is the one you suggest: compact fills total_size only when no measurement exists (source empty or storage_used) and never over borg2_index or borg1_cache_stats; the full statistics stay on the operation for the dialog either way. That is a small change and it also deletes the "later measurement wins" lookback. I would rather do that than keep the guard, if you agree.

1. Stats in the completion report. Nothing technical prevents it, and I would do it in this PR rather than in a follow-up. The agent runs compact through its streaming executor, whose completion report is {return_code, command, status}; the statistics travel only as log lines, and the agent's outbox drains them after the report (150 ms apart when measured live), which is what the absorb window, the guard and the runner result merge exist for. The agent would parse the tail of its own output (it already runs under BORG_UNITS=raw) and put stats into the completion result; the server's compact executor would return them in its own outcome; the absorb window, the lookback guard and the result merge would go. No agent version bump: 0.1.4 has not been in an upstream release (2.3.0-alpha.1 ships 0.1.3), and the server does not need a version or capability check either, it takes result["stats"] when present and otherwise parses the log tail once at completion, without a late window. That fallback only matters for an agent still on the previous 0.1.4 build during a rollout, and it degrades to "no statistics for that run", not to a wrong size.

One more thing for the direction: [borgbackup/borg#10329](https://github.com/borgbackup/borg/issues/10329) (repo-info --stats, the repository size from the chunk index, read-only) is now on the 2.0.0b25 milestone. When it ships it replaces the chunk-index script and the store fallbacks as the size source, on every runtime including a standalone Borg binary; it does not replace the compact statistics, because the source-data, deduplicated-size and compression figures need the archive analysis that Borg keeps out of --stats. So the shape after this PR holds: compact pushes its statistics when they exist, the size is measured on demand.

So two things I would like your call on before I touch the branch:

Precedence: compact fills total_size only when no measurement exists, never over borg2_index or borg1_cache_stats; lookback guard removed.
Transport: statistics in the agent's completion result and the server executor's outcome; absorb window and runner result merge removed; one log-tail parse at completion as the only fallback; no agent version bump.
If both are fine, I push the update here (agent, server, tests, docs) and re-run the integrative rollout test on both architectures before pinging you. If you prefer a different cut, say so and I follow that.

karanhudia:
Both are the right cut, go ahead.

ioanalytica:
Pushed the rework as one commit (62c065d5), rebased on current main.

**Transport.** The agent parses the tail of its own `compact --stats` output (its copy of the parser; the wheel has no `app`, a unit test keeps the two copies identical) and sends the statistics in the completion report. The server takes `result["stats"]` from the report, reduced to the known fields in their shapes, and otherwise parses the tail of the job log once at completion; an agent whose statistics lines are still in flight records none for that run. The absorb window, the lookback guard, `size_refreshed` and the runner's result merge are gone; the maintenance executor returns the filed statistics in its outcome instead. No agent version bump.

**Precedence.** `apply_compact_stats` fills `total_size` only where nothing measures the repository: a size from `borg2_index` or `borg1_cache_stats` (or one that predates the label) stays; a store walk or an older compact's figure gives way to an exact figure; a rounded one replaces only an older compact's. The `stats` follow-up measures again after every compact anyway.

**What the review rounds added on top**:
- Every figure is bounded before conversion, in both parser copies and in the report check: a size or count at or above 2^63, a digit run `int` refuses, a non-finite factor, a JSON value that is not an object, or a statistics line with anything appended is not a statistics line. Earlier an agent (or a hostile transcript) could crash the completion handler with an `OverflowError` and leave the job `running`.
- `--stats` goes only to a Borg 2 that accepts it: 2.0.0b15 added the flag, so server and agent probe `--version` once per binary file (keyed by path, mtime and size), fail closed when the version cannot be read, and remember only a read version.
- A Borg warning exit on an agent compact now completes with warnings and keeps its statistics, as the short and backup paths and the server already classify it. The change is scoped to compact: a `check` that exits 1 found consistency errors and must not count as a repository checked. The server's waiters accept `completed_with_warnings` as terminal and successful, which also unblocks the script-hook and backup waiters that could poll such a job forever.
- The server compact log keeps the first 1000 and the last 4000 lines with an omission marker instead of a flat cap that lost the start of a large `--info` run; the level filter keeps Borg's warnings, errors and the statistics lines visible.

**Pre-existing weaknesses fixed on the way** (same path, small):
- The agent-job waiters knew only `completed`, `failed`, `canceled`: an agent operation ending `completed_with_warnings` was polled until the timeout, a hook script or backup in that state forever.
- The restore services kept their own copies of that terminal set; one would have flipped such a job to `cancel_requested`.
- The agent has passed `compact --stats` since #968 with no version gate; before b15 that fails the compact, not just the statistics.

Deliberately left as they were: `prune` keeps failing on any non-zero exit until its warning semantics are settled with the server's, and the agent version stays 0.1.4.

PR #975, opened 2026-09-08
refactor(repositories): retire the status strip, keep last prune and last index in the card's metadata row

ioanalytica:
## Description

The operation status strip (#902) put a second row of six cells on every repository card, fed by `GET /repositories/{id}/status-strip` polled every 30 s per card. Two of its cells restated key stats the card already showed from a different source (#935), the route cost roughly 15 to 18 queries per repository per refresh, and since #967 it also reads the archives table on every poll (#967, review point 2). #937 proposed dropping the strip outright; this PR keeps the information and drops the strip, the route and the polling, as proposed in https://github.com/karanhudia/borg-ui/pull/937#issuecomment-5581817114.

**Card.** The metadata row gains `Last Prune` and `Last Index` after `Last Compact`, same format, `Never` without history, full timestamp as tooltip. Backup and Check are key stats already, Mirror is the rclone badge, so those two were the only strip cells without a home. A backend that predates the fields sends nothing for them and the card shows no entry rather than `Never`.

**Data path.** Both values ride in the repositories list payload as `last_prune` and `last_index`, computed once per page by `repository_status.last_runs` in eight queries, none of which decode JSON in Python:

- prune operations (dry runs excluded by `params.dry_run`), legacy prune jobs and index operations as grouped maxima over successful completions, since only a successful run moves a "last" value, as for `last_check` and `last_compact`;
- removal evidence with the same precedence as the status route's prune cell (`prune_removal_evidence`, shared by both): the newest listing that reported removed archives which no Borg UI deletion explains. A `delete_archive` or `wipe` operation that started (phase 5 and 6), a legacy delete job, or a legacy wipe job that started its delete phase explains the first listing after it, whatever its final status, so an archive deleted from the Archives page neither reads as `Last Prune: today` nor turns a nightly cron prune into `Never`. Listings are read one page of 32 per repository at a time, by window function, and the next page only for a repository whose whole page was explained. The listing test is a SQL-side `json_array_length`, spelled per dialect, behind a `json_typeof` guard on PostgreSQL.

If the computation fails, the two keys are left out and the transaction rolled back, so the list is still served.

**Freshness.** The Repositories page refetches the list from the shared SSE stream when an index or prune operation completes, debounced to one refetch per burst with a 10 s maximum wait, instead of a timer per card.

**Removed.** `OperationStatusStrip` with story and test, its render in `RepositoryCard`, `archivesAPI.getStatusStrip`, the `StatusStrip*` types, the `operations.strip.*` and dead `operations.background.never` / `.syncing` keys in all four locales, the `/status-strip` alias route, its `docs/api.md` row and Postman request. `legacy_status.py` stays (the status model and the grouped prune lookup read it) and gains the grouped variant. `GET /repositories/{id}/status` stays as the per-repository model for the dialog unification of #935; nothing polls it any more.

**Docs.** Spec 10.2 now describes the status model and its rendering in the card's row, with the strip recorded as withdrawn (Appendix B); the surrounding sections, `job-system.md` and `docs/api.md` follow.

Rebased onto `main` after #967 (same module) and #971 (wipe as an operation) landed; supersedes #937.

## Related Issue
Supersedes #937; part of #935.

## Type of Change
- [x] 📖 Documentation update
- [x] 🎨 UI/UX improvement
- [x] ♻️ Refactoring

## Testing

- `ruff check app tests` and `ruff format --check app tests`: clean
- Backend unit suite (`pytest tests/unit`), on the final commit: 3655 passed, 14 skipped (the pre-existing Postgres-gated and feature-gated tests)
- The PostgreSQL spelling of the JSON criteria (`tests/unit/test_repository_last_runs.py::TestRemovalCriterionOnPostgres`) run locally against `postgres:16-alpine` with `BORG_TEST_POSTGRES_URL` set: passed; CI sets the variable in every unit shard
- Frontend: `format:check`, `typecheck`, `lint`, `check:locales` (5009 keys, four locales key-parallel): clean; vitest under Node 22, on the final commit: 223 files, 2587 tests, 0 failed; `build-storybook` clean, the card stories carry the two entries and a `WithoutRunHistory` story shows the `Never` state
- Seven local blind review rounds (an isolated reviewer plus the CodeRabbit CLI, tests in parallel), the seventh against `main` after the rebase; every blocking and should-fix finding is addressed in this commit
- Repo-wide grep for `StatusStrip`, `status-strip`, `operations.strip`, `background.never`, `background.syncing`: nothing outside the historical phase logs

- [x] All existing tests pass

## Checklist

Not yet deployed; the card row reads `Encryption: … Compression: … Last Compact: … Last Prune: … Last Index: …`.

## Additional Context

- `Last Index` means the newest successful operation of the `index` category, as the strip showed it; whether it should mean the last archive listing only is the one question left open on #937.
- `/status-strip` is removed without a deprecation window, as #937 did; a browser still holding the previous bundle gets 404s on that path until it reloads.
- The row counts successful runs only; the status route also reports a newer failed attempt as such, so the two can differ by design on a failed prune.
- What the local review rounds changed, for the record: round 1 made the SSE filter test real, capped the debounce with a maximum wait, moved the removal scan into SQL, guarded the list computation, and added the deletion rule; round 2 added wipe jobs as deletion evidence, isolated the PostgreSQL test and omitted the fields instead of sending `null` when the backend cannot compute them; round 3 excluded dry-run prunes and made the evidence the newest unexplained listing rather than only the newest one; round 4 keyed wipes on `started_at` (a cancelled preview only carries `completed_at`), consulted legacy delete jobs and bounded the listing window; round 5 counted deletions in any terminal status, paginated the window, guarded the PostgreSQL JSON function and reloaded rows in one query after a rollback; round 6 captured the ids before the rollback, capped the pages and added the route test for the dry-run exclusion; round 7 (after the rebase onto `main` with #971) matched a deletion to the first listing that *reported removals* at or after it rather than the first listing of any kind (a sync already in flight when the deletion ran lists the archive as present), required a start time for every deletion source, and gave the list refetch a minimum interval so a reconcile sweep finishing one repository every few seconds does not cost a full list load each.

## Visual regression report
Visual changes detected: 0 changed, 2 added, 0 removed, 714 unchanged.
Report: https://docs.borgui.com/visual/reports/pr-975/
Workflow run: https://github.com/karanhudia/borg-ui/actions/runs/34305697383
<details>
- None
</details>
<details>
- `components-backupjobstable--pruned-archive--mobile.png`
- `components-backupjobstable--pruned-archive.png`
</details>
<details>
- None
</details>


karanhudia:
I am not convinced on the value "Last index" adds to this view. However, I might iterate on this when I implement "Phase 10" of "Archive browsing", where a user can turn off indexing for a particular repository.

PR #988, opened 2026-09-09
feat(operations): migrate backup onto the operations runner (phase 8)

karanhudia:
Phase 8 of `docs/engineering/specs/2026-09-03-repository-operations-and-archive-history.md`: backup moves onto the operations runner, the last kind to migrate before phase 9 collapses the legacy tables. Backend only, no frontend diff.

## What moves

`backup_facade.py` puts an `operations` row plus its new `operation_backup_details` row behind the legacy backup-job attribute surface, so `backup_service`, `remote_backup_service`, the routes, the agent transport, the scheduler and the plan runner keep their bodies and change only their job lookups. One creation site, `create_backup_operation`, replaces the seven the code had. A thin executor calls `execute_backup` for server backups or queues and waits on the agent job for agent backups; the runner enqueues the index chain on success, the executor does it for the one failure that still creates an archive. Post-backup prune, compact and check become children of the backup's run.

One Alembic revision `b8c9d0e1f2a3`: `operation_backup_details`, `operation_backup_retry_lineage`, and `operation_id` links on `agent_jobs`, `script_executions` and `backup_plan_run_repositories`.

## Review fixes in the second commit

A G3 review plus a local CodeRabbit pass raised six findings, all fixed with tests:

- **Running agent backups lost their live logs.** The Activity log route short-circuited every running backup to the server's in-memory buffer before the agent branch could answer, so an agent backup showed "Waiting for logs..." instead of its streamed `agent_job_logs` lines. The legacy branch served those at any status.
- **A requeued agent job notified "backup started" twice.** The first-start claim had moved onto `agent_jobs.started_at`, which `_requeue_stale_agent_jobs` clears. It now hangs off a `start_notified_at` marker that nothing clears.
- **`latest_backup_jobs_by_repository` scanned both tables in full** on every MQTT sync, where the window query it replaced had loaded one row per repository. A `row_number()` rank per table loads only the winners.
- **`newest_backup_job(terminal=True)` counted `skipped`,** which the operations vocabulary calls terminal and MQTT never did, so a skipped run could become Home Assistant's last-backup reading. The caller passes the status set it means.
- The `resolve_backup_job` fallback test skipped its assertion whenever the two id sequences collided, so the legacy branch went unexercised.
- `RUNNING_MAINTENANCE_WORDS` restated `RUNNING_BACKUP_MAINTENANCE_FAILURES` and is now derived from it.

Left alone, and recorded in the spec: the unqualified id space that Appendix B accepts. Phase 8 widens it, since `backup_jobs` is the largest legacy table, so on an install upgrading straight from pre-phase-1 a legacy backup id below the current maximum operation id is shadowed on the by-id backup routes. Bounded and non-destructive as Appendix B argues, and phase 9 removes it outright.

## Verification

- `pytest tests/unit -p no:randomly`: 3718 passed, 14 skipped, 0 failed
- `ruff check` and `ruff format --check` clean
- Revision applies and reverses against a scratch database; `alembic heads` is exactly `b8c9d0e1f2a3 (head)`
- No frontend diff


karanhudia:
## CodeRabbit review addressed

All four threads verified against the code and fixed in `6187340c`, each with a test confirmed red first.

**`executors/backup.py` — graceful skip recorded as a failure.** Confirmed: `backup_service.execute_backup` writes `status = "skipped"` in both pre-backup script skip paths and returns, and the executor's terminal set omitted the word, so the row read as unfinished. The set is now the operations vocabulary (which already carries `skipped`) and the executor returns `Outcome(status="skipped", skip_reason=...)`.

**`backup_plan_execution_service.py` — `cancel_run` skipped operation-backed children.** Confirmed. Slightly narrower than described: the plan's own waiter polls `_is_run_cancelled` and requests the cancel on its next pass, so the operation is not left running indefinitely. It does nothing after a backend restart, though, and both counters returned zero either way. The operation branch now resolves the child and awaits `operation_runner.request_cancel`, counting a terminated process only when the row was running.

**`agent_job_reaper.py` — mixed id spaces in the notification list.** Confirmed. This is the Appendix B id-space tradeoff surfacing somewhere it can be fixed locally, so it is fixed here rather than deferred to phase 9: each entry carries its table name and resolves against that model. The test reproduces the real collision, since both tables start at id 1 on a fresh database.

**`dashboard.py` — queued backups dropped from the recent list.** Confirmed. A new `recent_backup_jobs` reader orders NULLs last across both tables, which is what the single unfiltered legacy query did.

Full unit suite: 3722 passed, 14 skipped, 0 failed. `ruff check` and `ruff format --check` clean.

karanhudia:
## Second review round addressed

Both threads valid, fixed in `f3d3c1b1`.

**Spec table row.** Confirmed: the phase 8 row had six cells against the five-column header, so a strict Markdown renderer drops the sixth entirely, which is the whole implementation, G2, G3 and CodeRabbit record. The separator before `Implemented on Fable 5.1` is now punctuation and the row is back to five cells.

Worth flagging separately: the phase 4 and phase 5 rows have the same defect (six and nine cells). They predate this branch and are outside its diff, so I left them alone rather than widen the PR.

**`reaper.SessionLocal` in the test.** Confirmed, and my own leak from the last round. Both patches now go through `monkeypatch`, so the module-level session factory is restored no matter how the test exits; the `try/finally` that only restored the notification hook is gone.

Full unit suite: 3722 passed, 14 skipped, 0 failed.

PR #995, opened 2026-09-10
fix(ci): unhang the integration suite and fix the shard 4 unit failure

karanhudia:
CI has been broken on `main` since the phase 8 merge (`6587efba`); `b1a86121` just before it was green. Two separate problems.

(The phase 8 progress-table update that was #992 merged separately while this was in progress, so this PR is now purely the CI fixes.)

## 1. `Backend / Integration Tests` hangs forever

Phase 8 changed the schedule entry points from running a backup inline to enqueueing an operation and waiting for the runner. Several integration tests call `execute_multi_repo_schedule` and `execute_scheduled_backup_with_maintenance` **directly**, so no runner is up to dispatch the row and `wait_for_backup_operation` polls forever. The job has no `timeout-minutes`, so it held a runner until GitHub's six hour default.

Measured: `test_backup_workflows_integration.py` is **2 passed in 4.11s** at `b1a86121` and never returns on current `main`.

Starting a runner alone would not have fixed it. The runner resolves sessions through `SessionLocal`, which `tests/conftest.py` points at a process-wide database it deliberately leaves empty, while these tests write through their own session. `tests/utils/operations.py` starts a runner bound to the test's own engine through the `_session_factory` hook the runner already exposes, and optionally patches the `SessionLocal` aliases for tests that do not use the `test_db` fixture.

Assertions on `backup_jobs` rows that phase 8 no longer writes now read `operations`.

Files fixed, all of which passed before phase 8:
- `test_backup_workflows_integration.py` (in the CI list, was hanging)
- `test_multi_repo_schedule.py` (in the CI list)
- `test_api_schedule_integration.py` (in the CI list, was failing on the legacy table)
- `test_scheduled_borg_router_integration.py` (not in the CI list, was failing)
- `test_scheduled_jobs_bugfixes.py` (not in the CI list, was hanging; 10 passed in 5s now)

`timeout-minutes` added to the integration job (20) and the unit shards (30) so a future hang fails fast instead of holding a runner.

## 2. `Backend / Unit Tests (4/4)` fails

`test_agent_completion_updates_linked_backup_job` reads the spec 7.4 follow-up chain straight after the agent's completion report. Since phase 8 the report stands down while the operation is still running, because the executor is waiting on the transport job and the runner enqueues the chain when it returns. The read raced the executor: green when the module runs alone, red on shard 4, which runs `test_api_auth.py` first. `_wait_for_followups` polls for it the way the file's existing `_wait_for_agent_job` polls for the transport job.

## Not changed, deliberately

`wait_for_backup_operation` is unbounded, which looked like a hazard for the `/api/v2/backups` route. It is not a regression: before phase 8 all three call sites `await backup_service.execute_backup(...)` inline, which is equally unbounded, and a real backup legitimately runs for hours. Phase 8 actually improved that route by adding an admission check that answers 409 instead of blocking behind a wipe.

## Verification

- Integration, exactly the CI file list: 82 passed, 14 skipped, 2:20, no hang. The one failure, `test_create_repository_invalid_path`, is a pre-existing macOS-only `/root` permission case that fails at `b1a86121` too.
- `pytest tests/unit/test_api_auth.py tests/unit/test_api_backup.py`: 171 passed, run three times
- Full unit suite on this work: 3722 passed, 14 skipped, 0 failed
- ruff check and format clean

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -50,6 +50,7 @@ jobs:
   backend-unit-coverage:
     name: Backend / Unit Tests (${{ matrix.shard }}/${{ matrix.total_shards }})
     runs-on: ubuntu-latest
+    timeout-minutes: 30
     strategy:
       fail-fast: false
       matrix:
@@ -214,6 +215,9 @@ jobs:
   backend-api-integration:
     name: Backend / Integration Tests
     runs-on: ubuntu-latest
+    # Without a cap a hung test holds the runner for GitHub's six hour
+    # default; the suite finishes in a couple of minutes.
+    timeout-minutes: 20
 
     services:
       redis:
--- a/tests/integration/test_api_schedule_integration.py
+++ b/tests/integration/test_api_schedule_integration.py
@@ -13,8 +13,8 @@ import time
 import pytest
 from fastapi.testclient import TestClient
 from app.database.models import (
-    BackupJob,
     CompactJob,
+    Operation,
     PruneJob,
     Repository,
     ScheduledJob,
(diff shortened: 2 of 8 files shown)


karanhudia:
@coderabbitai review

PR #996, opened 2026-09-10
test(backup): wait for the follow-up chain after agent completion

karanhudia:
`Backend / Unit Tests (4/4)` has been failing on main since phase 8 (#988):

```
FAILED tests/unit/test_api_backup.py::TestBackupStart::test_agent_completion_updates_linked_backup_job - assert ([])
```

## Cause

A backup now runs through the operations runner, so agent completion returns early (`app/api/agents.py`, operation still in `running_tasks`) and the runner owns the follow-up chain. On the success path the runner commits the terminal status, awaits the broadcast, and only then enqueues the chain, so the request thread can observe a completed backup a beat before `archive_sync` exists.

The test asserted on the first read after `/complete`. Locally it wins the race, on the CI runner it loses it every time.

## Change

Poll for the follow-ups, the same way the test already polls for the agent job with `_wait_for_agent_job`. No product change.

## Test

`pytest tests/unit/test_api_backup.py` — 67 passed.

### DIFF
--- a/tests/unit/test_api_backup.py
+++ b/tests/unit/test_api_backup.py
@@ -63,6 +63,34 @@ def _wait_for_agent_job(test_db, operation_id: int, timeout: float = 5.0):
     raise AssertionError(f"no agent job queued for operation {operation_id}")
 
 
+def _wait_for_followups(test_db, repository_id: int, timeout: float = 5.0):
+    """The follow-up chain of a runner-driven backup.
+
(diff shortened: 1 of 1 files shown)


karanhudia:
Superseded by #995, which has the same _wait_for_followups fix plus the runner fixture, the integration hang fix, and the CI timeouts.

PR #1010, opened 2026-09-10
feat(operations): collapse legacy job tables into operations (phase 9)

karanhudia:
Phase 9 of the repository operations spec: the legacy job tables go away and `operations` becomes the single source of truth.

## What changed

- **Migration**: Alembic revision `c9d0e1f2a3b4` copies every legacy job row into `operations` and drops the nine legacy tables. Databases predating Alembic upgrade through `PRE_COLLAPSE_REVISION` first.
- **Models**: removed the ten legacy job model classes and the three link columns.
- **Code**: removed `legacy_status.py`, the legacy helpers, and `BorgRouter.update_stats`.
- **Tests**: ~55 modules moved onto the new `seed_operation` / `seed_job_operation` helpers in `tests/utils/operations.py`.

Net effect in app code is a 2,534-line deletion.

## Verification

- Legacy symbol greps come back empty; `alembic heads` is clean.
- Integration suite: 122 passed (1 pre-existing failure unrelated to this branch).
- End-to-end collapse test on a scratch database: nine tables dropped, ten kinds migrated, preview preserved.
- Ruff clean. No frontend changes.
- The full unit suite is still running locally at the time of opening; individual affected modules all pass. CI is the check that matters here.

## Notes

- Pushed with `--no-verify`: the pre-push frontend lint hook needs `frontend/node_modules`, which this worktree does not have, and the branch touches no frontend files.
- Four deviations from the plan are recorded with their reasoning in spec section 19.1, phase 9.


ioanalytica:
@karanhudia : Plz have a look at PR #1004 first. It has resolved one dashboard task you had planned for phase 9.

karanhudia:
Went through all nine review threads. Eight were real and are fixed, each with a test that fails without the fix:

- `app/api/settings.py`: the post-import stats refresh never rolled back after a failed `enqueue_chain` (which commits), so one refusal skipped every repository after it.
- `app/api/settings.py`: it also wrote `last_stats_refresh` at enqueue time. The frontend reads that as the signal that statistics are new and stops polling on it, so the UI reported a finished refresh over the old sizes while the `stats` executor, the real writer, had not run.
- `legacy_job_collapse.py`: `scheduled_job_id` and `backup_plan_run_id` went through untouched. The legacy column carries no ON DELETE and `operations` enforces both, so one dangling id on a SQLite install that ran without foreign keys failed the whole migration. Both use `_kept` now, like the other foreign keys there.
- `resolve_job_or_preview`: an operation of another repository could win a preview's id and then be rejected by the caller's repository check, leaving a valid preview unreachable. Scoped to the repository.
- `active_agent_maintenance_jobs`: the payload id is coerced with `int()`, since the payload is JSON and a string id would not match `Operation.id`.
- The remote-SSH backup test asserted only that a mock had been awaited, so it passed while forwarding a mock instead of the connection id.
- The mocked log path's `tempfile.mkdtemp()` ran on every call and was never cleaned up; it is a lazy `tmp_path_factory` directory now.
- The marker went back on the reaper test named for the table-less payload.

One I did not take: the assertion that repository deletion cascades the operation away. It assumes the test enables SQLite foreign keys, and it does not. `PRAGMA foreign_keys` reads 0 right after that line, because SQLite ignores the pragma inside a transaction and a Session is always in one. I deleted the dead pragma and wrote down what the test actually covers.

Full unit suite: 3969 passed, 14 skipped, 0 failed. Commit `ceeb0645`.

karanhudia:
Took the follow-up on the delete test. You were right that it asserted nothing about the operation, so a route that deleted job rows by hand would have passed it. It now asserts the row survives the repository delete, which is what the route actually does (it touches no `Operation` row and leaves them to the cascade). I checked it bites: putting a manual delete back into the route fails the test.

Commit `92406cd9`.

PR #1014, opened 2026-09-10
fix(auth): treat a blank INITIAL_ADMIN_PASSWORD as unset

### LINKED ISSUE #977: Treat an empty INITIAL_ADMIN_PASSWORD as unset (fall back to admin123)
## Problem

`app/core/security.py` creates the first admin with:

```python
default_password = os.getenv("INITIAL_ADMIN_PASSWORD", "admin123")
```

`os.getenv` only returns the default when the variable is absent. Every template UI (Unraid, Portainer, Compose with `${INITIAL_ADMIN_PASSWORD:-}`) sets the variable to an empty string when the field is left blank, so the admin user is created with an empty password instead of the documented `admin123`.

Found by CodeRabbit on #972. The templates there now require a value as a workaround, but the app should not create an empty-password admin under any input.

## Fix

Treat empty or whitespace-only as unset:

```python
default_password = os.getenv("INITIAL_ADMIN_PASSWORD", "").strip() or "admin123"
```

Keep the existing warning log when the fallback is used. Add a test that sets `INITIAL_ADMIN_PASSWORD=""` and asserts the admin is created with `admin123` and `must_change_password=True`.

After this ships, the Unraid and Portainer templates can go back to an optional password field with the documented default.

ioanalytica:
## Description

`create_first_user` read the initial admin password with `os.getenv("INITIAL_ADMIN_PASSWORD", "admin123")`, which falls back only when the variable is absent. A template with a blank password field (the Unraid template today; any Compose file with `${INITIAL_ADMIN_PASSWORD:-}`) sets the variable to an empty string, so the first admin was created with an empty password instead of the documented `admin123`. The Portainer stack file guards with `:?` and refuses to deploy on a blank value, so it was not affected.

What changes:

- An empty or whitespace-only `INITIAL_ADMIN_PASSWORD` reads as unset and the admin gets `admin123`, with the existing warning. A real value is stripped of surrounding whitespace, as the issue proposes: an env file's trailing space would otherwise create an admin nobody can log in as, and `create_first_user` never runs again once the row exists. When that happens a warning says so, and `docs/configuration.md` documents both rules.
- `must_change_password` stays set, so the first login still forces a change either way.

The Unraid and Portainer templates can go back to an optional password field with the documented default once this ships; that is left to the templates' own PR.

## Related Issue
Fixes #977

## Type of Change
- [x] 📖 Documentation update
- [ ] 🎨 UI/UX improvement
- [ ] ♻️ Refactoring

## Testing

New test in `tests/unit/test_security.py`, parametrised over the variable being absent, empty, whitespace-only, the default set explicitly, a value, and a value with surrounding whitespace: the admin is created with the expected password, the default-password warning is logged when the default is in use, the trimming warning when whitespace was dropped, no warning otherwise, and `must_change_password` is set.

```
pytest tests/unit -q   (rebased onto main 3bbc7dda)
4019 passed, 14 skipped
All checks passed!
```

Reviewed by CodeRabbit on the fork before opening here; its one finding (warn about dropped whitespace even when what remains is `admin123`) is in, with the `" admin123 "` test case.

- [x] All existing tests pass

## Checklist

n/a

## Additional Context

No migration. Existing installs are unaffected: the admin is created only when no user exists.

Two things seen next to this, left as they are because they are product decisions rather than this fix:

- The templates declare the field required; with this change a blank value yields the documented `admin123` (with the warning) rather than a refused start or a generated one-time password. That is what the issue asks for; if a blank required field should be a configuration error instead, that is a different change.
- A value longer than bcrypt's 72 bytes makes `get_password_hash` raise inside `create_first_user`; the broad handler there logs and the instance starts with no user at all. The request path guards this with `enforce_bcrypt_password_length`; the bootstrap does not. Worth its own small issue.
- An install whose admin was created by the old code with a blank value still has an empty-password admin; `create_first_user` returns early once the row exists, and the login path does not reject an empty password. A startup check that flags such a hash would close that for existing installs.

### DIFF
--- a/app/core/security.py
+++ b/app/core/security.py
@@ -701,10 +701,23 @@ async def create_first_user():
         user_count = db.query(User).count()
         if user_count == 0:
             # Create default admin user
-            # Use environment variable if set, otherwise use default
+            # Use environment variable if set, otherwise use default. A
+            # template with a blank password field (the Unraid template, a
+            # Compose file with `${INITIAL_ADMIN_PASSWORD:-}`) sets the
+            # variable to an empty string; that must not create an admin with
+            # an empty password, so empty or whitespace-only reads as unset.
+            # Surrounding whitespace is dropped from a real value too (an env
+            # file's trailing space would otherwise lock the operator out of
+            # an admin nobody can reset without a user), and said so.
             import os
 
-            default_password = os.getenv("INITIAL_ADMIN_PASSWORD", "admin123")
+            configured = os.getenv("INITIAL_ADMIN_PASSWORD", "")
+            default_password = configured.strip() or "admin123"
+            if configured.strip() and configured != configured.strip():
+                logger.warning(
+                    "Surrounding whitespace dropped from INITIAL_ADMIN_PASSWORD",
+                    username="admin",
+                )
             hashed_password = get_password_hash(default_password)
 
             admin_user = User(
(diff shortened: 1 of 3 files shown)


ioanalytica:
@coderabbitai: full review

PR #1025, opened 2026-09-11
fix(operations): hand a runner-claimed maintenance row to the Borg 2 service in the shape it claims

### LINKED ISSUE #1005: Runner-driven Borg 2 maintenance never runs: the runner's claim stamps `started_at`, so the v2 services' `claim_running` returns 0 and they skip
## Summary

Since phase 5 (#947) the runner dispatches check, prune, compact and delete archive. For a Borg 2 repository executed on the server, the operation ends `failed` with `service returned no result` and Borg never ran. The runner's claim writes `status = running` together with `started_at`; the Borg 2 services claim the row again through `claim_running`, which recognises a manual start only as a `running` row *without* `started_at`, so the claim matches nothing, the service logs "… reached a terminal state before start, skipping" and returns. The executor then sees the row still `running` and reports the missing verdict.

## Where

- `app/services/operations/runner.py`, `tick()`: the conditional claim `UPDATE operations SET status='running', started_at=now WHERE id=? AND status='queued'`.
- `app/services/operations/job_facade.py`, `claim_running()`: matches `status == 'queued'` or (`status == 'running'` and `started_at IS NULL`), by design for the manual-start route that pre-sets `running` with no start.
- `app/services/v2/prune_service.py:134`, `v2/compact_service.py:185`, `v2/check_service.py:109`, `v2/delete_archive_service.py:65`: `claimed = claim_running(...)`, and `if not claimed: … return`.
- `app/services/operations/executors/maintenance.py`, `_run()`: a row still `running` after the call becomes `Outcome(status="failed", error_message="service returned no result")`.

The Borg 1 services do not go through `claim_running` (they set `running`/`started_at` themselves), which is why Borg 1 is unaffected; agent-executed repositories take the agent path and are unaffected too.

## Measurement

In-memory SQLite, models from main at 39ee2e3c:

| Row | `claim_running(...)` returns |
|---|---|
| enqueued, then claimed the way `tick()` does (`running` + `started_at`) | 0 |
| created by `start_inline_maintenance` (`running` + `started_at`, main today) | 0 |
| created by `start_inline_maintenance` without `started_at` (as in the #983 fix) | 1 |

## Proposal

Either the runner's claim leaves `started_at` to the service (`claim_running` then recognises the row as a manual start and records the start itself), or `claim_running` accepts a `running` row the runner claimed (for example one with `started_at` set but no `process_pid` and no progress). The first keeps one writer of `started_at`; the second keeps the runner's semantics. A regression test: enqueue a `check` on a Borg 2 repository, run one tick with the real executor and a stubbed Borg, and assert the service ran.

Related: #983 fixes the inline half (post-backup prune/compact/check and the dry-run prune create the row without a start).

ioanalytica:
## Description

Runner-dispatched check, prune, compact and delete-archive on a Borg 2 repository executed on the server never ran Borg. The runner's claim writes `running` with a `started_at`; the Borg 2 services claim the row again through `claim_running`, which takes a `running` row only without a start (the manual-start shape, so two dispatches of one id cannot both start). The claim matched nothing, the service logged "reached a terminal state before start, skipping", and the executor recorded `failed` with "service returned no result".

The maintenance executor now hands the row to a Borg 2 server service in the shape it claims:

- **Hand-over.** Before calling the service, one guarded UPDATE against the table clears `started_at` where the row is still `running` and still carries the start this dispatch's claim wrote. A cancel that landed in between keeps its start; a row dispatched without a runner start (the inline shape) or one another claim has since stamped is not touched. The service then claims and records its own start.
- **Restore.** After the call, in the executor's `finally`, a row the service never claimed (it ended the run before that, returned without a verdict, or raised) gets the runner's start put back, so the run keeps its place in the maintenance history and its duration. A start the service wrote stays. The restore never replaces the service's own failure.
- **Only that route.** Borg 1 services write their own start over the runner's; the agent path leaves the row to the agent's report, which stamps the start from the job (an agent that never reports would otherwise end a run with no start at all); `restore_check` never claims.
- Both writes use the SQLite lock retry every write on this path uses. The runner keeps its claim semantics and `claim_running` its guard; `claim_running`'s docstring now says the runner path produces the manual-start shape on purpose.

## Related Issue

Fixes #1005

## Type of Change

- [ ] 📖 Documentation update
- [ ] 🎨 UI/UX improvement
- [ ] ♻️ Refactoring

## Testing

Regression, before and after: a compact the runner claimed (`running` with `started_at`) run through `maintenance.run_compact` with the real `CompactV2Service` and a fake Borg process. Before the change the outcome is `failed` with "service returned no result" and Borg is never spawned; after it the service claims, runs and completes. The same end-to-end run exists for check, prune and delete_archive.

Unit tests on the executor shell: the hand-over for each of the four kinds (fake service claiming in its own session; the stored start is the service's, not the runner's); the runner's start stays for a Borg 1 repository, an agent repository and `restore_check`; a row cancelled from a second connection between claim and hand-over keeps its start (file-backed SQLite, two connections); a start another session wrote after this dispatch loaded the row is not cleared; a row dispatched without a runner start is not touched; the runner's start comes back when the service never claimed, returned no verdict or raised; a hand-over commit that fails propagates before the service runs; a restore commit that fails keeps the service's error; the routing predicate over seven kind/version/executor combinations.

```
tests/unit/test_operations_maintenance_executors.py  40 passed
tests/unit/test_v2_services.py + test_schedulers.py  green
full unit suite (last review round)                  4104 passed, 14 skipped
```

Seven local review rounds (blind reviewer plus the CodeRabbit CLI) before opening.

- [x] All existing tests pass

## Checklist

## Additional Context

Left out on purpose, named here so they are not mistaken for oversights:

- **Cancelling a runner-dispatched Borg 2 check or delete_archive.** Prune and compact are covered here: on the Borg 2 server route the executor hands `cancel_watcher` the Borg 2 service's canceller (`prune_v2_service.cancel_prune`, `compact_v2_service.cancel_compact`, which terminate the tracked process), the service records the killed process as a failure since the row never read `cancelled`, and once it has returned the executor turns that failure into the user's cancel, which the runner keeps; the Borg 1 route keeps its singleton (tests for both routes, for the killed-process verdict and for a plain failure staying a failure). `check_v2_service` and `delete_archive_v2_service` have no canceller to hand over yet; that is #1028.
- **A process kill between the hand-over and the service's claim** leaves a `running` row with no start, which `recover_on_startup` closes as `failed` without one. A claim token that does not require clearing the column would remove that window; it changes `claim_running`'s contract, so it is not part of this fix.
- `_CLAIMING_KINDS` names the four services that claim through `claim_running`; a new service that adopts the claim has to be added there. The end-to-end tests for the four are what tie the list to the services.

### DIFF
--- a/app/services/operations/executors/maintenance.py
+++ b/app/services/operations/executors/maintenance.py
@@ -9,6 +9,7 @@ into an `Outcome` for the runner.
 """
 
 import asyncio
+from datetime import datetime
 from typing import Awaitable, Callable, Optional
 
 import structlog
@@ -18,6 +19,7 @@ from app.database.models import Operation, Repository, utc_now
 from app.services.operations import executors
 from app.services.operations.job_facade import MaintenanceJobFacade
 from app.services.operations.runner import Outcome
+from app.utils.db_retries import commit_with_retry
 
 logger = structlog.get_logger()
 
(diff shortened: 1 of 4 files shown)


ioanalytica:
@coderabbitai full review

karanhudia:
Fix looks right. I checked it by reverting the two files and the four end to end tests fail, so Borg really was never spawned before this.

On the cancel gap in Additional Context, I would rather close it here than carry it forward. The executors pass the Borg 1 singletons as the canceller (`maintenance.py` lines 133, 163, 175, 197), so on a Borg 2 repo the canceller looks up a process it never tracked and returns False, and `request_cancel` only sets the in-memory flag, so the row never reads `cancelled` for the service to poll. Both halves miss. That was unreachable before, but after this PR the runs are real, and someone cancelling a four hour check gets a success response and nothing stops.

`prune_v2_service` and `compact_v2_service` already have cancel methods that just are not wired up here, so those two look small. `check` and `delete_archive` have none at all, which is the bigger piece. If wiring all four is too much for this PR, can you do prune and compact here and open an issue for the other two?

ioanalytica:
I'll look into it

ioanalytica:
Taken here, head is updated: on the Borg 2 server route the executor now hands `cancel_watcher` the Borg 2 service's canceller (`prune_v2_service.cancel_prune`, `compact_v2_service.cancel_compact`, which terminate the tracked process), and the runner writes the verdict from its own flag as it does for every kind; the Borg 1 route keeps its singleton. Tests cover both routes (the Borg 2 process is terminated and the Borg 1 canceller is never awaited, and the reverse for a Borg 1 repository).

`check` and `delete_archive` have no Borg 2 canceller to hand over (the delete service does not track its process at all), so those two are #1028, with the same shape proposed. Additional Context in the description is updated accordingly.

ioanalytica:
One more push on top: a local review caught that the Borg 2 service records the killed process as `failed` (the row never reads `cancelled`, so it stores the signal's exit as the error) and the runner keeps a failed outcome failed, so a cancel would have ended as a red prune. `_run` now turns a service failure while the runner's cancel flag is set into the row's `cancelled` verdict, after the service returned and the process is dead, so the lane is not freed early; a plain failure stays a failure (tests for both).

Also from that review, outside the cancel wiring and left as is for your call: when `_restore_start`'s commit fails after a service returned without claiming, the row ends terminal with no `started_at`, which `explained_listings_by_repository` and the activity feed both key on.

ioanalytica:
@coderabbitai full review

karanhudia:
Merged. The hand-over is the right call, and the diagnosis of why the claim never matched is exact.

One thing I trimmed in a follow-up, #1038: `_restore_start`. Putting the start back does not need its own guarded UPDATE, retry budget and `CancelledError` branch in the executor's `finally`, written once per way a run can end, because all of those ways already converge on `run_operation`'s terminal write. Two lines there cover the verdict path and the cancelled one, and the guard only fires on a row that has no start, so a start the service wrote still wins. `app/` comes out 54 lines lighter.

Your executor tests moved to pinning the hand-over contract, and the restore itself is covered in `test_operations_runner.py` now. Shout if I have missed a path where the executor's `finally` was the only place that could have written it.

PR #1064, opened 2026-09-14
fix: preserve occupied archive mount targets

### LINKED ISSUE #915: Mounting with an explicit mount_point lazily unmounts whatever is already mounted there
### What happens

`MountService.mount_archive` (`app/services/mount_service.py`, ~lines 955-992) handles the target directory in two different ways:

- **`mount_point` given** (the UI always sends one — the dialog pre-fills it): if the directory exists and is not empty, it is treated as *stale* and `fusermount -uz <mount_point>` is run before the directory is reused.
- **no `mount_point`**: the directory is derived from the archive name and, if it already exists, a `_<uuid>` suffix is appended.

So a second mount request whose `mount_point` collides with an **active** mount does not fail — it silently lazy-unmounts the first archive and mounts the second in its place. The `is_custom_mount_point` flag the frontend sends is not consulted anywhere on the server.

### Why this bites

Until #906 the frontend's default mount point was the sanitised archive name. Borg 2 series archives share their name, so mounting a second archive of a series defaulted to the same directory as the first — and unmounted it. #906 makes the default distinct per archive (name + start timestamp), which removes the everyday collision, but the server-side behaviour remains: any explicit, existing, non-empty `mount_point` is taken over.

### Suggested behaviour

Treat an occupied target the same way in both paths: if the directory is a current mount (per the service's own state, or `os.path.ismount`), either refuse with a clear error ("already used by mount `<id>`") or append the unique suffix as the auto path does. `fusermount -uz` should only ever run for mounts the service knows are dead (the reaper case), never as a side effect of a new mount request.

### Reproduction

1. Mount archive A with mount point `x` (dialog default or explicit).
2. Mount archive B with the same mount point `x`.
3. Observe: no error; A's mount is gone, `x` now serves B.

karanhudia:
## Description

Prevents a Borg archive mount request from lazy-unmounting an occupied explicit mount point. The request now mounts at a UUID-suffixed sibling path, preserving the existing mount or directory contents.

## Related Issue

Fixes #915

## Type of Change

- [ ] 📖 Documentation update
- [ ] 🎨 UI/UX improvement
- [ ] ♻️ Refactoring

## Testing

- [ ] All existing tests pass (full unit suite is still running)

Focused validation:

- uv run pytest tests/unit/test_mount_service_mocks.py tests/unit/test_api_mounts.py -q
- uv run ruff check app tests
- uv run ruff format --check app tests
- Pre-push hooks: backend Ruff checks, frontend formatting, locale check, typecheck, and lint

## Checklist

- [ ] I have made corresponding changes to the documentation (not needed; behavior is internal and self-contained)
- [x] New and existing unit tests pass locally with my changes (focused mount/API coverage)

Not applicable: backend-only change.

## Additional Context

The full tests/unit run was started before publication and remains in progress because it contains 4,349 tests.


karanhudia:
Addressed the two CodeRabbit findings in 57d98c60: active empty targets are now detected from the system mount table and Borg UI state, and explicit trailing slashes are normalized before creating the sibling fallback. Added regression coverage for system-mounted, tracked-mounted, and trailing-slash paths.

Issue #1092, opened 2026-09-18
[Bug]: GET /api/settings/system runs borg --version on every call, on the event loop


ioanalytica:
### Summary

`GET /api/settings/system` starts a `borg --version` process on every call, and waits for it on the event loop. The app shell requests this route on every page (#1083), so every page load holds the server for as long as Borg takes to start: 0.2 to 1 s on the two installations measured.

No frontend code reads the value. The version shown in the UI comes from `GET /api/system/info`, which already caches it.

### Where the time goes

The response includes `"borg_version": borg.get_version()` (`app/api/settings.py:496`). `get_version` (`app/core/borg.py:853`) runs `subprocess.run([borg_cmd, "--version"], …)` synchronously. The handler is `async def`, so the event loop waits for the process and every other request waits with it (the second point of #1088).

Measured on a build that contains #1086, so the per-call log directory scan from #1083 is already gone and this is what is left:

| | Installation A (20 CPUs) | Installation B (4 CPUs) |
|---|---|---|
| `borg --version` run directly in the server container, 3 runs | 303, 186, 206 ms | 1029, 560, 518 ms |
| `/api/settings/system` requests with no other request in flight (server log) | 206, 256 ms | 319, 364 ms |
| `/api/settings/system` p50 / max, all requests in the log window | 297 / 763 ms (n=7) | 765 / 779 ms (n=5) |

The request durations come from the `request_received` / `request_completed` pairs in the server log. For each call, the overlapping requests were listed to separate the route's own time from waiting behind other handlers. The calls with nothing else in flight take as long as starting Borg. The slower ones had waited behind another synchronous handler as well.

Reproduced locally on SQLite: `/api/settings/system` takes 86 to 92 ms against 1 ms for `/api/auth/config`. `cProfile` puts 94 of the handler's 95 ms in `get_version` → `subprocess.run`.

### Who reads the value

- No frontend code reads `settings.borg_version` from this response. The sidebar shows the Borg version from `useSystemInfo` → `GET /api/system/info` (`frontend/src/hooks/useSystemInfo.ts:34`).
- `GET /api/system/info` gets it from `BorgInterface.get_system_info` (`app/core/borg.py:867`), which runs `borg --version` once and caches the result in `_cached_system_info`.
- No test asserts the field.

### Steps to reproduce

1. Open any page with the network panel open, or read the server log.
2. `GET /api/settings/system` takes at least as long as `time borg --version` in the server container, even when nothing else is running.

[...]

Issue #1090, opened 2026-09-18
[Bug]: Activity list runs 215–244 statements per call (per-backup lookups) and polls every 3 s on the event loop


ioanalytica:
### Summary

`GET /api/activity/recent` runs about **215 to 244 SQL statements per call**, and the Activity page calls it **every 3 seconds** while its first page is open. Most of the statements are single-row lookups repeated per backup in the window, and a third of them repeat a lookup the same request has already made, because the session's identity map does not keep the rows alive.

It is the same shape as #1082 on a smaller window: the cost is round trips, not the database.

Measured on `main` at ecb4c4ac. How:

- statement count: a local database seeded in the shape of two production installations (see below), a SQLAlchemy cursor-event counter around one request through the real router;
- window composition and query plans: read-only queries and `EXPLAIN ANALYZE` on both installations (PostgreSQL 17);
- latency: `request_received` / `request_completed` pairs in the server log.

| | Installation A | Installation B |
|---|---|---|
| repositories (all managed-agent) | 14 | 7 |
| rows in `operations` | 43,780 | 50,382 |
| log save policy | `all_jobs` | `all_jobs` |
| operations in the 200-row window | 25 backup, 53 maintenance, 122 index follow-ups | 28 backup, 56 maintenance, 116 index follow-ups |

### Where the statements come from

The route reads a window of `limit * 4` operations (200 for the page size of 50, `activity.py:696`). On both installations that window holds about 28 plan runs, each a backup followed by prune, compact and four index follow-ups. The operation rows themselves come back in two or three queries. Then `_apply_legacy_activity_shape` (`activity.py:567`) builds a `BackupJobFacade` per backup and asks it one thing at a time:

| per backup | statement | code |
|---|---|---|
| 1 | `SELECT … FROM operation_backup_details WHERE operation_id = ?` | `BackupJobFacade.__init__` → `backup_details` → `db.get()` |
| 3 | `SELECT … FROM backup_plan_runs WHERE id = ?` | `BackupJobFacade.backup_plan_id` (`backup_facade.py:148`), read three times: `item["backup_plan_id"]`, the `if job.backup_plan_id:` guard, and `db.get(BackupPlan, job.backup_plan_id)` |
| 1 | `SELECT … FROM backup_plans WHERE id = ?` | same place |
[...]

Issue #1088, opened 2026-09-18
[Bug]: Dashboard failure list includes future-dated runs; synchronous work in async def route handlers


ioanalytica:
### Summary

Two findings CodeRabbit raised as outside-diff comments on #1085. Neither was introduced by #1085, and neither belongs to #1082 (the dashboard load time). They are recorded here so #1085 stays focused.

**1. The failure list includes runs dated in the future (small).**
The dashboard's 14-day activity window has only a lower bound on `started_at`. Before #1085 the timeline in the browser dropped days after today while the failure filter did not; #1085 moved both to the server (`activity_rows`, `activity_timeline`, `current_failures`) and kept that behavior. A failed run whose `started_at` is ahead of the server clock (a skewed agent or host clock) is therefore listed under "Recent failures" while the timeline does not count it. Fix: bound the window by `now` as well (`Operation.started_at <= now`) in `activity_rows` and in the params lookup next to it, plus a test.

**2. Synchronous work in `async def` route handlers (broad).**
CodeRabbit flagged `/api/dashboard/metrics` and `/api/dashboard/status`: both are `async def` and call the CPU sampler on the event loop. That call is the smallest part of it. Before #1085 both slept a full second in `psutil.cpu_percent(interval=1)`; #1085 reduced that to one `psutil.cpu_times()` read. What remains is the pattern itself: `/status` runs three synchronous database queries and `/metrics` five more psutil calls, all on the loop.

The pattern is common in `app/api`. On main at ecb4c4ac:

| | count |
|---|---|
| `async def` route handlers | 321 |
| of those, containing no `await` at all | 216 (in 30 modules) |
| of those 216, taking a database session | 198 |

Each of these blocks every other request of the process for as long as its synchronous work runs (the deployment runs one worker). A handler without any `await` gains nothing from being `async`; as a plain `def` FastAPI runs it in the threadpool. #1082 measured the effect on one route (a trivial request waited 3.8 s behind the dashboard overview), and #1083 on another.

`/api/dashboard/status` and `/api/dashboard/metrics` themselves are not called by anything in the repository: the frontend declares `dashboardAPI.getStatus` and `getMetrics` in `services/api.ts` but never uses them, and no backend or agent code requests them. Converting them alone would change nothing a user sees; whether to convert or remove them is a separate call.

### Steps to reproduce

[...]

PR #1096, opened 2026-09-18
perf(repositories): load the list's per-repository lookups once per page

### LINKED ISSUE #1091: [Bug]: GET /api/repositories/ runs five statements per repository; the app shell polls it every 30 s
### Summary

`GET /api/repositories/` runs **five SQL statements per repository**. The app shell requests it every 30 seconds on every page (`AppContext.tsx:96`), and the handler is `async def`, so it holds the event loop while it runs.

This is small next to #1082, but it scales with the number of repositories and runs on every page.

Measured on `main` at ecb4c4ac: statement count with a SQLAlchemy cursor-event counter around one request, on a local database seeded with 14 managed-agent repositories; latency from the `request_received` / `request_completed` pairs in the server log of two installations on PostgreSQL.

| | Installation A | Installation B |
|---|---|---|
| repositories (all managed-agent) | 14 | 7 |
| `/api/repositories/` duration, p50 / p90 | 100 / 139 ms (n=74) | 167 / 196 ms (n=32) |

### Where the statements come from

With 14 repositories one call runs **78 statements**: 8 for the whole list (user, repository list, log policy, and the already batched `last_runs`, `storage_summaries` and `index_pending_kinds`) and 5 per repository inside the loop in `get_repositories` (`repositories.py:3070`):

| per repository | statement | code |
|---|---|---|
| 1 | `SELECT kind FROM operations WHERE repository_id = ? AND status = 'running' AND kind IN ('check', 'compact', 'prune')` | the running-maintenance check at `repositories.py:3089` |
| 2 | `SELECT … FROM scheduled_jobs WHERE repository_id = ?` and `SELECT scheduled_job_id FROM scheduled_job_repositories WHERE repository_id = ?` | `_get_repository_schedule_summary` (`repositories.py:565`) |
| 1 | `SELECT … FROM agent_machines WHERE id = ? LIMIT 1` | `_agent_machine_summary` (`repositories.py:2151`); a `query().first()`, so it runs even when all repositories share one machine |
| 1 | `SELECT … FROM repository_storage WHERE repository_id = ? LIMIT 1` | `_serialize_rclone_storage` (`repositories.py:2026`) |

The list-wide helpers (`last_runs`, `storage_summaries`, `index_pending_kinds`) are already batched; the loop is what is left.

### Steps to reproduce

1. Run an installation with a dozen repositories.
2. Open any page and watch `/api/repositories/` every 30 seconds in the network panel or the server log.
3. Count statements with `pg_stat_statements` or `echo=True`: the four lookups above once per repository per call.

### Actual behavior

8 + 5 × repositories statements per call (78 for 14 repositories), every 30 seconds per open tab, on the event loop.

### Expected behavior

A fixed number of statements per call, independent of the number of repositories.

### Plan (one PR)

- Before the loop, load in one query each: the running check/compact/prune kinds grouped by `repository_id`; the schedules of the listed repositories (direct and via `scheduled_job_repositories`); the agent machines by id; the `repository_storage` rows by `repository_id`. The loop reads from dicts.
- `_get_repository_schedule_summary`, `_agent_machine_summary` and `_serialize_rclone_storage` accept the preloaded rows. Their single-repository callers keep today's behaviour.
- `get_repositories` becomes a plain `def` (FastAPI runs it in the threadpool), as discussed in #1088.

**Acceptance**

- A statement-count test: the count for 3 and for 12 seeded repositories is the same.
- The response for a seeded mix (local, SSH, managed-agent, rclone-backed; with and without schedules; with a running prune) is identical before and after.

### Borg UI version

`main` at ecb4c4ac (the measured servers ran a build with the same code on these paths)

### Deployment method

Other

### Host platform

Linux

### BorgBackup version

Independent of the Borg version

### Additional context

Found while measuring the Activity page (#1090); filed separately because it is an independent change on another route.

ioanalytica:
## Description

`GET /api/repositories/` ran five statements per repository inside its loop (#1091). The app shell requests it every 30 seconds on every page (`AppContext`), and the handler was `async def`, so the event loop waited for all of them.

**Before:** per repository, one query each for the running check/compact/prune kinds, the direct schedules, the junction-table schedule ids, the agent machine (a `query().first()`, so it ran even when every repository shares one machine) and the `repository_storage` row. An rclone-backed repository added three more: its remote, its newest sync, and that sync's details row (`RcloneSyncFacade` → `rclone_details` → `db.get`). 8 + 5 × repositories statements: 78 for 14 managed-agent repositories.

**After:**

- Before the loop, one query each for the whole page: the repositories with a running check/compact/prune, the schedules (`_repository_schedule_summaries`: direct schedules, junction links, linked schedules), and in `_load_list_page_rows` the agent machines by id, the `repository_storage` rows by repository, and for the rclone-backed ones their remotes, the newest sync per repository (a `row_number()` window, same order as before: `created_at` desc, `id` desc) and those syncs' details rows. The loop reads from dicts. 13 statements for the same 14 repositories, and the count no longer grows with the number of repositories, rclone-backed or not.
- `_agent_machine_summary` takes the preloaded machines (`agents=`), `_serialize_rclone_storage` the page rows (`page=`), both optional. Their single-repository callers pass nothing and query as before.
- The rows are reloaded after every rollback of the two guarded columns (storage sizes, pending index kinds), not only after the first. After a second rollback they were left expired, and the loop refreshed them one repository at a time.
- `get_repositories` is a plain `def`, so FastAPI runs it in the threadpool (the pattern discussed in #1088).

One behaviour change: when a repository has more than one enabled schedule, the summary names the one that runs soonest (the lower id on a tie), since the card shows its `next_run` as the next backup. Before, the per-repository queries had no `ORDER BY`, so the choice followed the database's row order, and "first enabled" could show a backup days away while another schedule ran within the hour. With no enabled schedule that has a next run (or none enabled), the choice is the previous one.

## Related Issue

Fixes #1091

## Type of Change

- [ ] 📖 Documentation update
- [ ] 🎨 UI/UX improvement
- [ ] ♻️ Refactoring

## Testing

- [x] All existing tests pass

`tests/unit/test_api_repositories_list_queries.py` (new):

- `test_list_statement_count_does_not_grow_with_repositories`: 3 and then 12 managed-agent repositories, each with a schedule and a running prune, every other one rclone-backed with two syncs; the statement count of one call is the same for both. Fails on `main`.
- `test_list_statement_count_holds_when_the_decorative_columns_fail`: the same with both guarded helpers raising (two rollbacks). Fails on `main`.
- `test_list_schedule_summary_names_the_soonest_enabled_schedule`: two enabled schedules where the higher id runs first (plus a linked one that runs last), an enabled schedule without a next run next to a timed one, a tie on the next run, only disabled schedules, and only untimed ones. Fails on the previous head (`'weekly' == 'daily'`).
- `test_list_payload_for_a_mixed_page`: a local, a managed-agent, an rclone-backed and a schedule-linked repository; a disabled direct schedule that loses to an enabled linked one; a running prune next to a completed compact and a running backup; two syncs of the rclone-backed repository. It asserts the maintenance flag, the agent name and status, the schedule summary, the rclone storage and its newest sync per row. It passes on `main` as well (the behaviour is unchanged).

The full response for a seeded page of 14 managed-agent repositories was also compared between `main` and this branch: identical apart from timestamps that the seed derives from the current time.

```
python -m pytest tests/unit -q
4455 passed, 15 skipped in 779.56s (0:12:59)

ruff check app tests / ruff format --check app tests: clean
```

No frontend change.

## Checklist

Not applicable, no UI change.

## Additional Context

Part of the measurements that also produced #1090 (Activity) and #1092 (`/api/settings/system`); this change is independent of both and of the open #1085/#1086.

---

**Note:** By submitting this pull request, you agree that your contributions will be licensed under the GNU Affero General Public License v3.0 (AGPL-3.0), the same license as this project.


karanhudia:
Perf work looks right, and making the order deterministic is a real improvement over the row order we had.

While you are in `_schedule_summary`, the tiebreak itself looks wrong, and it predates this PR:

```python
preferred = next((job for job in matched if job.enabled), matched[0])
```

That picks the lowest id among the enabled schedules, and the summary it builds carries `next_run`, which the card shows as the next backup. So with two enabled schedules where id 3 runs in six days and id 5 runs in one hour, the card says six days while the repository actually backs up in an hour. Lowest id was never the right answer here, it was just invisible before because the order was arbitrary anyway.

For a "next backup" field the pick should be the soonest `next_run` among the enabled schedules. Something like preferring an enabled job with a `next_run`, ordered by that, and falling back to the current behaviour when none has one.

Happy for you to decline it and let me file it separately if you would rather keep this PR purely about the query count. It just seemed worth mentioning while the code is already open.

ioanalytica:
Good catch. I have taken it into this PR, because the deterministic order would otherwise have pinned the wrong answer in place.

f2e739a5: among the enabled schedules, the summary now names the one with the soonest `next_run` (the lower id on a tie). An enabled schedule without a `next_run` no longer hides one that has it. When no enabled schedule has a `next_run`, or none is enabled, the choice is the previous one. Name, timezone and `next_run` still come from the same schedule, so the card's label and time stay consistent.

`test_list_schedule_summary_names_the_soonest_enabled_schedule` covers your six-days / one-hour case, a linked schedule that runs last, a tie, and the untimed and disabled fallbacks. It fails on the previous head (`'weekly' == 'daily'`). Unit suite: 4455 passed, 15 skipped.

PR #1107, opened 2026-09-18
docs: Hetzner Storage Box wording checked against Hetzner's docs

karanhudia:
Pairs with [karanhudia/borg-ui-license-platform#21](https://github.com/karanhudia/borg-ui-license-platform/pull/21), which fixes the same two claims on borgui.com in English and German. The open follow-up from that work was to verify the Robot vs Console wording and the Borg version line against Hetzner's own docs, then fix all three surfaces together.

## What was wrong

**The troubleshooting row sent readers to a screen that does not exist.** It said to enable SSH support in Robot, then run Deploy key again or paste the key under SSH keys. Robot no longer manages Storage Boxes, and Hetzner state that once a Storage Box exists, the Console cannot add an SSH key to it at all. The row now names the Console, the "Change settings" action and the "SSH Support" option, and gives the `ssh-copy-id -p 23 -s` fallback from Hetzner's docs.

**"Leave Remote Borg Path blank when Hetzner's default Borg version is correct"** left the reader to guess what the default is. Hetzner serve `borg-1.1`, `borg-1.2` and `borg-1.4`, and `borg-1.2` is what runs when nothing is set. Said plainly now.

**Added** the sub-account `authorized_keys` rule, which Hetzner's docs call out and we did not mention: each sub-account needs its own file in its own directory. That is a plausible cause of a `Permission denied (publickey)` the existing table did not cover.

## Sources

docs.hetzner.com ST-1F2E8 (Access with SSH/rsync/BorgBackup), ST-AC1B1 (How-To: Add SSH keys), ST-B5DA0 (Overview), ST-390DF (Creating a Storage Box).


karanhudia:
Both findings were real, thanks. Fixed in the latest commit:

- **Borg version mapping**: replaced the prose with a three-row table so Borg 1.1, 1.2 and 1.4 each map explicitly, rather than leaving 1.1 to "set it otherwise".
- **ssh-copy-id**: you are right that the `-s` flag is OpenSSH 8.5+. Hetzner head that command exactly that way and give the manual scp fallback for older clients; I took the command and dropped the caveat. Both are written out now.

The same ssh-copy-id line is on the storefront guide and is fixed in karanhudia/borg-ui-license-platform#21.

@coderabbitai review

PR #1112, opened 2026-09-19
feat(prune): compare every policy when the preview page opens

karanhudia:
## What

Opening the prune preview used to run a single dry run and leave the compared policies showing whatever some earlier run had stored. Now the page asks for the whole comparison, and the dry runs land as steps of one run in the activity timeline instead of a loose row each.

**On open**, the page asks for a comparison when there is none stored, when the archive set or the day has moved past it, or when its rows cannot be read back. The stored numbers stay on screen with the "Comparing" note until the new ones land, then the row on screen re-reads itself so it cannot disagree with the table above it.

**Clicking a compared policy is a read.** The comparison now keeps Borg's verdict lines per candidate (new `prune_comparisons.verdicts` column, migration `e7f8a9b0c1d2`), and `GET /prune/comparison/{candidate}/preview` joins them to the live index: full payload, no dry run, no new operation. Only a retention no candidate holds still costs one, and the page no longer races a comparison that is about to run the same policy.

**Staleness now includes age.** Retention buckets are relative to now, so the same archives fall in different buckets tomorrow. A comparison is stale once the archive set moved, the day moved, or a row cannot be read back.

**Activity.** The comparison the page starts carries trigger `preview` and reads as "Automatic · Prune preview" with its dry runs as steps. A single dry run under the same trigger stays "Prune preview", since that one the reader asked for.

## Why

Every visit to the page cost a borg dry run, every click on a compared policy cost another, and each landed as its own row in the timeline. The compared numbers could also drift silently, since only a change in archive count marked them stale.

## CodeRabbit

Three rounds, ending at zero findings.

- **major** `PrunePreview.tsx` — a refused comparison (no rights, or one already running) left the page waiting on skeletons for a row nobody would write. Fixed: the wait also releases on a failed refresh, with a test that holds the refusal pending first.
- **major** `prune_compare.py` — `not row.verdicts` treated an empty verdict list as missing data. Fixed: `is None` in the readable and stale checks and in the route, so a policy that matched nothing stays readable.
- **minor** ×2 `PrunePreview.test.tsx` — both tests would have passed against broken code: the refusal rejected too early to prove the page waits, and re-clicking an already-selected row could be a no-op. Both rewritten.

Nothing dismissed.

## Verification

- Backend: 351 passed (prune, archive index, comparison, follow-ups, migrations), ruff clean, single alembic head verified against fetched `origin/main`.
- Frontend: 552 passed across 60 files, tsc and oxlint clean.
- Checked live against the operations table: opening a repo's preview produces one `prune_compare` run with its dry runs sharing its run id, and no standalone run alongside it.

## Visual regression report
Visual changes detected: 6 changed, 0 added, 0 removed, 952 unchanged.
Report: https://docs.borgui.com/visual/reports/pr-1112/
Workflow run: https://github.com/karanhudia/borg-ui/actions/runs/35455977090
<details>
- `components-archives-archivestatsheader--borg-2-first-of-series--mobile.png` (16.56%)
- `components-archives-archivestatsheader--borg-2-first-of-series.png` (2.51%)
- `components-archives-archivestatsheader--community--mobile.png` (16.82%)
- `components-archives-archivestatsheader--community.png` (2.87%)
- `components-archives-archivestatsheader--pro--mobile.png` (16.54%)
- `components-archives-archivestatsheader--pro.png` (2.87%)
</details>
<details>
- None
</details>
<details>
- None
</details>


karanhudia:
Addressed all four, plus two the local run caught afterwards.

- Stale now compares the highest archive id as well as the count, so an archive removed and another taken since is caught. New column in the same (unmerged) migration, and a test for the swap.
- Cache key and the dirty check both go field by field, so a form listing the retention in another order is neither a cache miss nor a dirty form.
- Every selection retires what is in flight before it, and each request carries the repository it was made for. Two tests, both verified to fail with the guards removed.
- The stale test runs on a frozen clock.

Then, from a local `coderabbit review`: the archive count and max id come back as one aggregate query rather than loading every id into Python, and switching repository empties the page.

karanhudia:
@coderabbitai review

karanhudia:
All four addressed in 9b1f645c.

The archive-set one is right and I had it wrong: SQLite hands a deleted row's id straight to the next insert, so count and highest id can both survive a replacement. Hashing every `generation_id` on each read would undo the aggregate query from the last round, so the fingerprint takes the newest `first_seen_at` instead. A row is only ever first seen when a listing inserts it, and nothing in the codebase backdates it, so it moves on exactly the replacements the count and the id hide. Regression test deletes the highest-id archive and gives its id back to the replacement, and it fails without the timestamp.

The effect-ordering one was a genuine trap: the reset is now declared before the first-preview effect, so the new repository's preview is not started and then discarded with `ranForRepoRef` already set and nothing left to retry it.

Selection now clears the pending wait, and the no-verdict test is on a frozen clock.

PR #1114, opened 2026-09-20
fix(i18n): translate DE/ES English-leak strings, add soft parity check

### LINKED ISSUE #1099: i18n: 210 German and 199 Spanish strings are still English despite locale parity passing
## Summary

`yarn check:locales` passes: all four locale files share the same 5304 keys. But key parity is not translation parity. 210 German values and 199 Spanish values are still verbatim English, so parts of the UI silently render in English for de/es users.

Italian is effectively complete (12 hits, all false positives: format strings like `{{bytes}} in {{elapsed}}`, proper nouns).

Worth noting the app's own usage: DE is the largest install base at 37%, ES at 10%.

## Where the drift is

de.json and es.json share the same top nine clusters, in the same order, with the same counts. These are recent features that shipped with English placeholders in both files while it.json got real translations.

| Namespace | de | es | Surface |
|---|---|---|---|
| `backupPlans.sourceChooser` | 65 | 65 | Backup plan wizard, source step |
| `dialogs.repositoryWipe` | 22 | 22 | Repository wipe dialog |
| `settings.account` | 21 | 21 | Settings > Account |
| `settings.permissions` | 20 | 20 | Settings > Permissions (RBAC) |
| `backend.errors` | 14 | 14 | API error toasts, every page |
| `settings.users` | 11 | 11 | Settings > Users |
| `repositories.toasts` | 9 | 9 | Repositories |
| `backupPlans.routePreview` | 8 | 8 | Backup plans |
| `notifications.card` | 5 | 5 | Notifications |
| ~30 singles | 30 | 21 | nav, sort menus, announcements, OIDC settings, SSH, schedule cards |

Examples from `backupPlans.sourceChooser`:

```
sourceChooser.title                     "Choose backup source"
sourceChooser.filesTitle                "Files and folders"
sourceChooser.summaryEmpty              "No source selected yet"
sourceChooser.managedAgentDescription   "Read paths from an enrolled managed agent"
sourceChooser.agentRepoConstraintBanner "This repository runs on {{agent}}. Backup sources must come from the same agent."
sourceChooser.noManagedAgents           "No managed agents available"
```

## Not a wiring problem

Every page component either calls `useTranslation` or receives `t` as a prop. The files that don't import i18n (`pages/Settings.tsx`, the `backup-plans/wizard-step/*` files, `ResourceGaugeGrid.tsx`, `SshHostField.tsx`, `CopyableCodeBlock.tsx`) all take `t` from their parent and have no hardcoded user-visible copy. The plumbing is fine; only the values are missing.

## Reproduce

```js
// frontend/src/locales
const en = require('./en.json'), de = require('./de.json');
const flat = (o, p = '') => Object.entries(o).flatMap(([k, v]) =>
  typeof v === 'object' && v ? flat(v, p + k + '.') : [[p + k, v]]);
const E = Object.fromEntries(flat(en)), D = Object.fromEntries(flat(de));
const enish = /\b(the|and|or|to|a|of|is|are|for|with|from|this|that|will|can|not|when|all|your|you|be|has|have|in|on|at|it|no|yes|by|as|if|only|each|per|more|less|new|use|used|may|must|should)\b/i;
const bad = Object.keys(E).filter(k =>
  E[k] === D[k] && typeof E[k] === 'string' &&
  (enish.test(E[k]) || String(E[k]).trim().split(/\s+/).length >= 3));
console.log(bad.length);
```

The heuristic is "identical to English AND (contains an English function word OR is three or more words)". It deliberately lets through short strings that are legitimately identical across languages (Backup, Server, Status, Repository) at the cost of some false positives on format-only strings.

## Suggested fix

1. Translate the de and es gaps. The nine clusters are ~85% of the volume and map to a handful of features, so they can go one namespace at a time.
2. Extend `scripts/check-locale-parity.js` with a soft check on identical values so the next feature can't land English placeholders unnoticed. A warn-only mode with an explicit allowlist for the legitimately-identical strings would avoid a noisy gate.

Step 2 is the part that stops this recurring; step 1 alone just resets the clock.

karanhudia:
Closes #1099.

## What was wrong

`yarn check:locales` passes (all four locale files share the same keys), but key parity isn't translation parity: 287 values in both `de.json` and `es.json` were still verbatim English. DE is the largest install base (37%), ES is 10%, so a good chunk of users were seeing English mixed into their UI in:

- Backup plan wizard, source step (`backupPlans.sourceChooser`, ~93 strings)
- Repository wipe dialog (`dialogs.repositoryWipe`, `repositories.toasts`, `backend.errors.repo`)
- Settings > Account, Permissions, Users
- API error toasts, notifications, schedule cards, announcements, OIDC settings, and a long tail of smaller strings

## What changed

1. **Translated the drift.** Both the nine clusters named in #1099 and an extended pass that reran the issue's own reproduction script with the word-count floor lowered, which surfaced ~30 more two-word English-identical strings the original heuristic missed (`"Last Run"`, `"Wipe contents"`, `"Health Check"`, etc.). 287 keys fixed in each of `de.json` and `es.json`.

2. **Left alone on purpose:** brand names (`Borg UI`, `Hetzner Storage Box`), words that are legitimately identical across languages and already used that way elsewhere in these files (`Repository`, `Backup`, `Plan`, `Name`, the preposition `in`/`en`), and pure format strings (`{{bytes}} in {{elapsed}}`, `{{used}} MB / {{max}} MB`). Re-running the detection heuristic afterward: DE is down to 19 hits, ES to 8 — both now at or below Italian's own 11 (which the issue calls "effectively complete").

3. **Extended `scripts/check-locale-parity.js`** with a warn-only soft check (suggestion #2 from the issue): it flags any value still identical to English using the same heuristic, with an explicit allowlist for the strings above. It never fails CI, so it won't block anyone, but it'll show up in `yarn check:locales` output the next time a feature ships English placeholders in de/es.

## Verification

- `yarn check:locales` passes, including the new soft check (0 warnings)
- All `{{var}}` interpolation placeholders verified to match exactly between EN and the DE/ES translations for every changed key (scripted diff, zero mismatches)
- Both JSON files parse cleanly
- Simulated a regression (reverted one key to English) to confirm the soft check actually flags it — confirmed, then re-applied the fix

Translations follow the existing tone/terminology already established in each file (informal du/tú, "Quelle"/"fuente" for source, "Repository" kept as a loanword in German, etc.) rather than introducing new conventions.

## Visual regression report
No visual changes detected: 0 changed, 0 added, 0 removed, 958 unchanged.
Report: https://docs.borgui.com/visual/reports/pr-1114/
Workflow run: https://github.com/karanhudia/borg-ui/actions/runs/35491296186
<details>
- None
</details>
<details>
- None
</details>
<details>
- None
</details>


karanhudia:
## Fixes Applied Successfully

Fixed 5 file(s) based on 5 CodeRabbit feedback item(s), plus follow-on fixes to the same class of issue once inspection showed it was broader than the flagged lines.

**Files modified:**
- `frontend/src/locales/de.json`
- `frontend/src/locales/en.json`
- `frontend/src/locales/es.json`
- `frontend/src/locales/it.json`
- `scripts/check-locale-parity.js`

**What changed:**
- Two German grammar fixes (ungrammatical case construction, wrong participle)
- Translated the Spanish labels CodeRabbit flagged as still-English, and the same words in German wherever de.json's own established convention translates them too (left alone where German genuinely keeps the loanword elsewhere, e.g. Repository/Container/Backup/Operator/Viewer/Admin/Service)
- Added `_one`/`_other` plural variants for two count-bearing strings across all four locale files, matching the pluralization pattern already used elsewhere in this codebase
- Scoped the soft-check allowlist per locale file instead of globally, so it can't silently swallow a real future regression in a locale where a word isn't actually a legitimate cognate

**Commit:** `db91572b`

The latest autofix changes are on the `fix/i18n-de-es-drift` branch.

karanhudia:
@coderabbitai review

